PR #32, opened 2026-03-05
fix env vars vmm

motatoes:
Seems that env vars broke after we moved to firecracker from podman. Fix is to store them and ensure to load them for every command pass the right envs to exec() as well as override the values that are passed ruing the run command

also should handle hibernate and restore


breardon2011 (approved):
approved

PR #45, opened 2026-03-06
Fix deploy-server workflow step ordering

motatoes:
we should pull the secrets before restarting the service


breardon2011 (approved):
approved

PR #50, opened 2026-03-07
Replace /commands with session-based exec API

motatoes:
## Summary

- Replace fire-and-forget `POST /commands` with persistent, reconnectable exec sessions
- Sessions keep running independently of client connections with 1MB scrollback buffer replay on reconnect
- WebSocket binary protocol: `[stream-id byte][payload]` for stdout/stderr/stdin/exit/scrollback_end
- Full implementation across proto, agent, worker, HTTP/WS handlers, TypeScript SDK, Python SDK, and CLI

## What changed

**Backend (Go)**
- New gRPC RPCs in agent + worker protos: `ExecSessionCreate`, `ExecSessionAttach` (bidi stream), `ExecSessionList`, `ExecSessionKill`
- `ScrollbackBuffer` — thread-safe ring buffer with fan-out to subscribers
- Agent-side session manager with process group kill (`-pgid`), env layering, disconnect timers, 5-min exited session GC
- Host-side `ExecSessionManager` mirroring the `PTYManager` pattern
- WebSocket auth now accepts token as `?token=` query param (needed because browsers/Node.js WS can't set custom headers)
- Deleted `internal/api/commands.go`

**SDKs**
- **TypeScript**: `Exec` class with `start()`, `attach()`, `run()`, `list()`, `kill()` — `run()` wraps in `sh -c` for shell convenience
- **Python**: `Exec` class with `run()`, `start()`, `list()`, `kill()` using WebSocket binary protocol

**Deploy fixes**
- Use `vmlinux-docker-5.10.bin` kernel (has `CONFIG_VIRTIO_VSOCKETS` — the old `vmlinux.bin` was a minimal 4.14 kernel without vsock)
- Load `vhost_vsock` kernel module on host
- Worker advertises public IP in `connectURL` for external SDK access
- Security group opens port 8081 for direct worker access

## Test plan

- [x] `make build` passes
- [x] `npx tsc --noEmit` passes (TypeScript SDK)
- [x] TypeScript SDK test: 19/19 assertions pass (`sdks/typescript/examples/test-exec.ts`)
- [x] Python SDK test: 17/17 assertions pass (`sdks/python/examples/test_exec.py`)
- [x] Tested on deployed instance at `50.17.44.74:8080`: sandbox creation, exec sessions, streaming output, kill, file ops, env vars, cwd, exit codes

### DIFF
--- a/internal/agent/exec_session.go
+++ b/internal/agent/exec_session.go
@@ -190,13 +190,17 @@ func (s *Server) ExecSessionAttach(stream pb.SandboxAgent_ExecSessionAttachServe
 		return fmt.Errorf("exec session %s not found", sessionID)
 	}
 
-	// Increment attached clients
+	// Increment attached clients — check if session is already dying
 	sess.mu.Lock()
-	sess.attachedClients++
 	if sess.disconnectTimer != nil {
-		sess.disconnectTimer.Stop()
+		if !sess.disconnectTimer.Stop() {
+			// Timer already fired — session is being torn down
+			sess.mu.Unlock()
+			return fmt.Errorf("exec session %s is shutting down", sessionID)
+		}
 		sess.disconnectTimer = nil
 	}
+	sess.attachedClients++
 	sess.mu.Unlock()
 
 	defer func() {


internal/agent/exec_session.go
```diff
@@ -0,0 +1,362 @@
...
+	s.execMu.RUnlock()
+	if !ok {
+		return fmt.Errorf("exec session %s not found", sessionID)
+	}
+
+	// Increment attached clients
+	sess.mu.Lock()
+	sess.attachedClients++
+	if sess.disconnectTimer != nil {
+		sess.disconnectTimer.Stop()
```
breardon2011: we could check the return value of .Stop() to see if session already dying, otherwise might have a race condition if connecting to a dying session, etc.-> if sess.disconnectTimer.Stop() = false would be timer already fired, attach should fail.

PR #51, opened 2026-03-09
feat/claude sdk abstraction

motatoes:
- **Replace /commands with session-based exec API**
- **fix: VM network connectivity + exec SDK improvements**
- **add claude agent sdk**

### DIFF
--- a/internal/api/agent_session.go
+++ b/internal/api/agent_session.go
@@ -127,6 +127,7 @@ func (s *Server) sendAgentPrompt(c echo.Context) error {
 		return c.JSON(http.StatusServiceUnavailable, errSandboxNotAvailable)
 	}
 
+	id := c.Param("id")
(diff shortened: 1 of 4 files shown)


internal/api/agent_session.go
```diff
@@ -0,0 +1,214 @@
...
+
+	return c.JSON(http.StatusOK, agentSessions)
+}
+
+func (s *Server) sendAgentPrompt(c echo.Context) error {
+	if s.execSessionManager == nil {
+		return c.JSON(http.StatusServiceUnavailable, errSandboxNotAvailable)
+	}
+
+	sessionID := c.Param("sid")
```
breardon2011: We never check if the session belongs to the sandbox. 

Similarly in internal/worker/handlers.go (different binary)

breardon2011 (approved):
Approve - one note

PR #130, opened 2026-04-14
Default sandbox timeout to 0 (persistent, never hibernate)

motatoes:
## Summary

- Changes the default idle timeout from `300s` to `0` across TS SDK, Python SDK, CLI, server API, and worker gRPC. `0` now means **persistent — never auto-hibernate**; callers opt into auto-hibernate by passing an explicit timeout.
- The router already supported \`timeout == 0\` as "no auto-hibernate timer" — this PR stops the API/worker layers from coercing \`0 → 300\` so the semantic actually reaches the router.
- Adds a new docs page \`sandboxes/timeout.mdx\` explaining the default, how to opt in, idle-based behavior, and auto-wake on SDK invocation.

## Motivation

Users running always-on agents (e.g. a Claw that listens for Telegram messages) currently get hibernated every 5 minutes, defeating the purpose. Exposing a \`timeout=0\` semantic and making it the default matches the mental model of "my sandbox stays alive until I kill it."

## Behavior

| `timeout` | Behavior |
| --- | --- |
| omitted / `0` | Persistent. Sandbox runs until explicitly killed or hibernated. |
| `> 0` | Auto-hibernates after N seconds of idleness (rolling timer). |
| `< 0` | Normalized to `0`. |

Any SDK operation transparently wakes a hibernated sandbox — no caller changes needed.

## Files

- **TS SDK** (\`sdks/typescript/src/sandbox.ts\`): \`create()\` and \`wake()\` default \`timeout\` to \`0\`; added JSDoc.
- **Python SDK** (\`sdks/python/opencomputer/sandbox.py\`): \`create()\` and \`create_from_checkpoint()\` default \`timeout=0\`; updated docstrings.
- **CLI** (\`cmd/oc/internal/commands/{sandbox,checkpoint}.go\`): \`--timeout\` flag defaults to \`0\` with "(0 = never hibernate)" in help text.
- **Server API** (\`internal/api/sandbox.go\`): \`createSandbox\`, \`wakeSandbox\`, \`createFromCheckpointCore\` stop coercing \`timeout == 0 → 300\`. \`setTimeout\` endpoint now accepts \`0\` (rejects only negative).
- **Worker gRPC** (\`internal/worker/grpc_server.go\`): same coercion fix on \`CreateSandbox\`, \`ForkFromCheckpoint\`, \`WakeSandbox\` paths so the new semantic works in multi-worker deployments.
- **Docs**: new \`docs/sandboxes/timeout.mdx\` page + mint.json nav entry.

## Breaking change

Clients that relied on the implicit 300s default will now get persistent sandboxes by default. They should either:
- Upgrade and explicitly pass \`timeout: 300\` where the old default was intended, or
- Accept the new persistent default and rely on manual \`kill()\`.

Other reference docs still say "default 300" in ~20 files (param tables, examples). Left as a follow-up since most of them are param tables rather than code-correctness issues.

## Test plan

- [ ] \`go build ./...\` on Linux CI (darwin build was verified locally for the packages touched — firecracker/qemu pre-existing failures on darwin are unrelated)
- [ ] \`go vet\` and \`go test ./internal/api/... ./internal/sandbox/...\` pass
- [ ] \`tsc --noEmit\` passes on the TS SDK
- [ ] Manual: create a sandbox with default settings, leave idle 10 minutes, confirm \`status=running\`
- [ ] Manual: create a sandbox with \`timeout: 60\`, leave idle, confirm \`status=hibernated\` after ~1 min
- [ ] Manual: \`sandbox.setTimeout(0)\` on a running sandbox disables its timer

### DIFF
--- a/cmd/oc/internal/commands/checkpoint.go
+++ b/cmd/oc/internal/commands/checkpoint.go
@@ -149,7 +149,7 @@ func init() {
 	checkpointCreateCmd.Flags().String("name", "", "Checkpoint name (required)")
 	checkpointCreateCmd.MarkFlagRequired("name")
 
-	checkpointSpawnCmd.Flags().Int("timeout", 300, "Timeout in seconds")
+	checkpointSpawnCmd.Flags().Int("timeout", 0, "Idle timeout in seconds before auto-hibernate (0 = never hibernate)")
 
 	checkpointCmd.AddCommand(checkpointCreateCmd)
 	checkpointCmd.AddCommand(checkpointListCmd)
--- a/cmd/oc/internal/commands/sandbox.go
+++ b/cmd/oc/internal/commands/sandbox.go
@@ -224,7 +224,7 @@ var lsShortcut = &cobra.Command{
 func init() {
 	// sandbox create flags
 	for _, cmd := range []*cobra.Command{sandboxCreateCmd, createShortcut} {
-		cmd.Flags().Int("timeout", 300, "Timeout in seconds")
+		cmd.Flags().Int("timeout", 0, "Idle timeout in seconds before auto-hibernate (0 = never hibernate)")
 		cmd.Flags().Int("cpu", 0, "CPU count")
 		cmd.Flags().Int("memory", 0, "Memory in MB")
 		cmd.Flags().StringSlice("env", nil, "Environment variables (KEY=VALUE)")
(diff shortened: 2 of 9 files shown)


breardon2011 (approved):
approvd

PR #142, opened 2026-04-14
billing: 10x memory tier prices + grandfather existing orgs

motatoes:
## Summary
- **10× all six memory tier rates** (1GB–64GB). `TierPricePerSecond` values are multiplied by 10, and each `TierPriceKey` is bumped one version (five tiers go from bare → `_v2`, 8GB goes `_v2` → `_v3` since it was already bumped once).
- **Grandfather every currently-paying org.** Migration 022 adds `orgs.price_locked BOOLEAN DEFAULT FALSE` and locks every org that has a `stripe_subscription_id` at migration time. These orgs stay on whatever Stripe Price IDs their subscription items already reference.
- **`cmd/migrate-prices` respects the lock.** Default query now filters `WHERE price_locked = FALSE`; pass `--force` to override (tagged `[LOCKED-override]` in the per-item log output so it's obvious).

## Net behavior
| Cohort | Outcome |
|---|---|
| 6 currently-paying 8GB orgs (just migrated to `sandbox_8gb_v2`) | Stay on `sandbox_8gb_v2` — no change |
| Any other currently-paying orgs at other tiers | Stay on their current price IDs |
| New signups post-deploy | Subscribe to `sandbox_Xgb_v2` / `sandbox_8gb_v3` at the new 10× rates |
| Future `migrate-prices --tier=X` runs | Skip locked orgs automatically; only touch new customers |

## On deploy
1. Merge this PR.
2. Next server boot, `EnsureProducts()` creates the six new Stripe Prices (same stable meters; new Price IDs with the new metadata keys).
3. Migration 022 runs and locks existing paying orgs.
4. **Do not run `migrate-prices` afterward** — that's the whole point. The new rates only apply to new subscriptions created via `CreateSubscription`.

Note: `backfill-disk-overage` intentionally does NOT respect `price_locked` — it adds a new billing dimension at the current rate rather than swapping an existing item's price, so skipping locked orgs there would leave them with unpriced disk-overage meter events (silently free disk above 20GB).

## Test plan
- [ ] Migration 022 applies cleanly on staging; all orgs with a non-null `stripe_subscription_id` have `price_locked = TRUE`, everyone else is `FALSE`.
- [ ] On server restart, logs show `billing: created price for sandbox_1gb_v2 (...)` through `sandbox_64gb_v2` plus `sandbox_8gb_v3`, all attached to the existing stable meters (`sandbox_compute_sandbox_1gb` etc.).
- [ ] `go run ./cmd/migrate-prices --tier=4096 --dry-run` (default, no `--force`) reports 0 candidates on prod (all current 4GB subs are locked).
- [ ] `go run ./cmd/migrate-prices --tier=4096 --dry-run --force` shows the locked orgs tagged `[LOCKED-override]` but still in the plan.
- [ ] Create a test signup via the normal Checkout flow → confirm the new subscription items reference the `_v2`/`_v3` price IDs at the new rates.
- [ ] Verify an existing 8GB org's invoice preview in Stripe still uses `price_1TMDFb...` (the v2 \$35/mo price we set earlier today), not the new `_v3` price.

### DIFF
--- a/cmd/migrate-prices/main.go
+++ b/cmd/migrate-prices/main.go
@@ -8,6 +8,9 @@
 //   DATABASE_URL=... STRIPE_SECRET_KEY=... \
 //     go run ./cmd/migrate-prices --tier=8192 --live
 //
+// By default, orgs with orgs.price_locked=TRUE are skipped (grandfathered).
+// Pass --force to override and migrate locked orgs too.
+//
 // The command is idempotent: items already attached to the target price are
 // skipped. Usage already accrued this cycle stays at the previous rate because
 // we pass proration_behavior=none; metered events are matched to whichever
@@ -32,6 +35,7 @@ func main() {
 	tier := flag.Int("tier", 8192, "memory_mb tier to migrate (must be a key in billing.TierPriceKey)")
 	dryRun := flag.Bool("dry-run", true, "if true, prints what would change without calling Stripe")
 	live := flag.Bool("live", false, "must be set to actually mutate Stripe; overrides --dry-run")
+	force := flag.Bool("force", false, "also migrate orgs with price_locked=TRUE (off by default — grandfathered orgs stay on their current price)")
 	flag.Parse()
 
 	if *live {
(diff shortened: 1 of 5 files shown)


motatoes:
Rebased on top of main (includes #147 per-second correction).

Changes during rebase:
- **TierPriceKey suffixes bumped one more step** so EnsureProducts creates fresh Prices at the 10× rates instead of finding the `_v2`/`_v3` Prices that #147 just created:
  - `sandbox_1gb_v2` → `sandbox_1gb_v3`
  - `sandbox_4gb_v2` → `sandbox_4gb_v3`
  - `sandbox_8gb_v3` → `sandbox_8gb_v4`
  - `sandbox_16gb_v2` → `sandbox_16gb_v3`
  - `sandbox_32gb_v2` → `sandbox_32gb_v3`
  - `sandbox_64gb_v2` → `sandbox_64gb_v3`
- `TierPricePerSecond` unchanged (still 10× of per-second rates).
- Comment on `TierPricePerSecond` updated to reflect the 2026-04-15 rollout (post-correction).
- Migration 022 unchanged — `UPDATE orgs SET price_locked = TRUE WHERE stripe_subscription_id IS NOT NULL` now locks the six orgs we migrated onto the corrected `_v2`/`_v3` Prices earlier today, freezing them at the correct per-second rate.

Net effect on deploy:
- New `_v3`/`_v4` Prices created at 10× rates → future signups see them
- Six existing orgs stay on their current (corrected per-second) `_v2`/`_v3` Price IDs → grandfathered
- `migrate-prices` skips locked orgs by default; `--force` still migrates them

Still independent of #146 (hide dashboard cost) which can merge any time.

breardon2011 (approved):
approved

PR #146, opened 2026-04-15
billing: hide cost display, link to Stripe Billing Portal

motatoes:
## Summary
Stops the dashboard from computing cost estimates locally; Pro users get a button that opens the Stripe Billing Portal instead.

## Why
Cost estimates on the billing page were computed from `billing.TierPricePerSecond` — a hardcoded Go map. After [#142](https://github.com/diggerhq/opencomputer/pull/142) (10× pricing + grandfather existing orgs), the six grandfathered orgs would see dashboard estimates at the new 10× rates while their actual Stripe invoices stay at the old rates.

Rather than build a rate-lookup layer that tracks historical prices per-org, sidestep the whole class of drift bugs: **Stripe is the source of truth for billing; don't duplicate it locally.** The Stripe Billing Portal shows current-cycle usage, invoices, subscription state, and lets customers update their payment method — strictly more than we had before.

### Backend
- `StripeClient.CreatePortalSession(customerID, returnURL)`: wraps `billingportal/session.New`
- `POST /api/billing/portal`: returns a portal session URL, using the request `Referer` as the return URL
- `GET /api/billing/get`: drops the `currentUsage` field from the response. Still returns `plan`, `stripeCreditCents`, `maxConcurrentSandboxes`, `hasPaymentMethod`

### Frontend
- Removes the "Current Month Usage" card (hardcoded-rate cost estimates)
- Removes the "Pricing Table" card (hardcoded \$/sec display)
- Adds a "Usage & Invoices" card with a button that opens the Stripe Billing Portal
- Invoice history section (already pulls from Stripe) unchanged
- Removes `BillingTierUsage` type and `currentUsage` field from `BillingState`

## Test plan
- [ ] Free user: sees only the Plan card + Upgrade CTA. No portal button. No cost display.
- [ ] Pro user: sees Plan card, the new Usage & Invoices card with a working portal button, Promotion Code card, and Invoice history.
- [ ] Clicking "Open Stripe billing portal" redirects to Stripe; closing the portal returns to `/billing` on the dashboard.
- [ ] `/api/billing/get` response no longer contains `currentUsage` (verify no other callers consume it — grepped, only `Billing.tsx` used it).
- [ ] Pro user with no `stripe_customer_id` (edge case): portal endpoint returns 400; frontend shows the error gracefully.

## Order-of-operations with #142
Merge + deploy this **first**. Then #142 can ship any time with no dashboard inconsistency — there's no cost display to be wrong. If shipped the other way around, the six grandfathered orgs would see 10× inflated estimates until this lands.

### DIFF
--- a/internal/api/dashboard_billing.go
+++ b/internal/api/dashboard_billing.go
@@ -6,12 +6,10 @@ import (
 	"log"
 	"net/http"
 	"strconv"
-	"time"
 
 	"github.com/google/uuid"
 	"github.com/labstack/echo/v4"
 	"github.com/opensandbox/opensandbox/internal/auth"
-	"github.com/opensandbox/opensandbox/internal/billing"
 )
 
 // billingSetup initiates the upgrade flow: creates Stripe customer + Checkout session.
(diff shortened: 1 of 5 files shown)


breardon2011 (approved):
approved

PR #151, opened 2026-04-15
docs: no-sudo CLI installer script

motatoes:
## Summary
- Replace the fragile inline `curl ... -o /usr/local/bin/oc` one-liner (requires sudo, breaks when copy-pasted with escaped parens) with `scripts/install.sh` piped via `curl | bash`.
- Installer detects OS/arch, installs to `~/.local/bin` (overridable via `OC_INSTALL_DIR`), warns if that dir isn't on PATH.
- Updated `docs/introduction.mdx` and `docs/guides/agent-skill.mdx` to point at the new script.

## Test plan
- [ ] Run `bash scripts/install.sh` on macOS (arm64) and Linux (amd64); confirm `~/.local/bin/oc` is created and executable.
- [ ] Run with `OC_INSTALL_DIR=/tmp/oc-test bash scripts/install.sh`; confirm target dir is respected.
- [ ] Run on a shell where `~/.local/bin` is not on PATH; confirm the warning prints.
- [ ] Render the updated `introduction.mdx` and `guides/agent-skill.mdx` pages and verify the snippets display correctly.

### DIFF
--- a/docs/guides/agent-skill.mdx
+++ b/docs/guides/agent-skill.mdx
@@ -18,7 +18,7 @@ This works with Claude Code, Codex, Cursor, and any agent that supports the [Age
 The `oc` CLI must be installed and configured:
 
 ```bash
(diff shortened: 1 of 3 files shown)


breardon2011 (commented):
approve

PR #156, opened 2026-04-16
feat(billing): free-tier trial credits with auto-hibernate

motatoes:
## Summary

Replace the static 5-concurrent-sandbox cap for free orgs with a **$5 virtual credits** model:

- New free signups get 500 cents of trial credits (`orgs.free_credits_remaining_cents`), decremented per-second by the usage reporter
- When credits hit zero: all running sandboxes are **force-hibernated** (with checkpoint keys preserved) and new create/wake/preview-URL-wake is blocked with `402 Payment Required`
- After upgrading to pro: users can **wake their hibernated sandboxes** and create new ones immediately
- Pro upgrade path ($30 Stripe credit via `ApplyPromotionalCredit`) is **untouched**
- Free-plan memory/CPU/disk tier caps remain in place (4 GB / 1 vCPU / 20 GB)

## Changes

- **Migration 023** (`internal/db/migrations/023_free_credits_remaining_cents.up.sql`): new `free_credits_remaining_cents` column (BIGINT, DEFAULT 500), backfills existing free orgs, resets usage watermark to prevent stale-history over-deduction
- **UsageReporter** (`internal/billing/usage_reporter.go`): new free-tier deduction pass after Pro/Stripe pass — computes cost via `CalculateUsageCostCents`, calls `DeductFreeCredits`, triggers enforcement on ≤ 0. Added `recover()` so a panic in any tick can't silently kill the reporter goroutine
- **EnforceCreditExhaustion** (`internal/billing/enforcer.go`): lists running sandboxes per org, calls gRPC `HibernateSandbox` on each worker (30s per-sandbox timeout), then updates PG directly: `EndScaleEvent` + `UpdateSandboxSessionStatus("hibernated")` + `CreateHibernation` with checkpoint key so the wake endpoint works post-upgrade
- **API gates** (`internal/api/sandbox.go`): `createSandbox` + `wakeSandbox` return 402 when `plan='free' && free_credits_remaining_cents <= 0`; concurrent-sandbox cap dropped for free plan (credits are the natural limit)
- **CP proxy gate** (`internal/proxy/controlplane_proxy.go`): `wakeHibernatedSandbox` (preview URL auto-wake) now checks credits before waking — previously had no plan check at all
- **Store** (`internal/db/store.go`): `FreeCreditsRemainingCents` field on `Org`, registered in `orgColumns` + `scanOrg`, migration registered as version 26
[...]

PR #172, opened 2026-04-17
fix(billing): enforce sandbox cap on free plan + show trial credits in UI

motatoes:
## Summary

- **Restore concurrent sandbox limit for free plan** — PR #156 removed the cap for free users, relying solely on credit deduction via the usage reporter. But if the reporter isn't running (no `STRIPE_SECRET_KEY`) or scale events aren't recorded (known race), credits never get deducted and there's no limit at all. The concurrent cap now applies to all plans again.
- **Surface free trial credits in billing API** — `billingGet` now returns `freeCreditsRemainingCents` so the frontend can display it.
- **Show credit balance on billing page** — Free users see their remaining trial credits (e.g. "$5.00 free trial credit remaining") in green, or a red exhausted warning when credits hit zero. CTA updated to "Upgrade to Pro for an additional $30 free credit and larger machine sizes". Polls every 30s to stay current.

## Test plan

- [ ] Free plan user sees credit balance on billing page
- [ ] Credit balance updates as usage reporter deducts
- [ ] Exhausted state shows red warning + upgrade CTA
- [ ] Free plan user hits 429 at concurrent sandbox limit (default 5)
- [ ] Free plan user hits 402 when credits exhausted
- [ ] Pro plan unaffected (concurrent cap still applies, no free credits shown)

### DIFF
--- a/internal/api/dashboard_billing.go
+++ b/internal/api/dashboard_billing.go
@@ -80,10 +80,11 @@ func (s *Server) billingGet(c echo.Context) error {
 	// the Stripe Billing Portal instead of re-computing cost locally against a
 	// hardcoded rate table that would diverge for grandfathered orgs.
 	return c.JSON(http.StatusOK, map[string]interface{}{
-		"plan":                   org.Plan,
-		"stripeCreditCents":      stripeCreditCents,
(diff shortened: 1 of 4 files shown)


breardon2011 (approved):
Approved

PR #173, opened 2026-04-17
fix(billing): pre-create session to fix scale event race

motatoes:
## Summary

- **Root cause**: `RecordScaleEvent` in the worker's `CreateSandbox` gRPC handler calls `GetSandboxOrgID` which queries `sandbox_sessions` in PG. But the control plane only created that row *after* the gRPC call returned — so the worker never found the org ID and silently skipped recording. Result: 73 out of 78 free orgs have zero scale events, credits never deducted, effectively unlimited free usage.
- **Fix**: Pre-generate the sandbox ID on the control plane, create the session with `status = "pending"` before the gRPC call, then promote to `"running"` on success or mark `"failed"` on error. The worker now finds the pending row and records the scale event correctly.
- Adds `CreateSandboxSessionWithStatus` to the store
- Adds `pending → running` and `pending → failed` transitions to `UpdateSandboxSessionStatus`

## Test plan

- [ ] Create sandbox on free plan → verify `sandbox_scale_events` row exists with correct org_id
- [ ] Verify usage reporter deducts credits after next tick (5 min)
- [ ] Verify gRPC failure marks session as `"failed"` (not orphaned as `"pending"`)
- [ ] Verify `CountActiveSandboxes` still only counts `"running"` (not `"pending"`)
- [ ] Verify existing create flows (combined mode, checkpoint fork) unaffected

### DIFF
--- a/internal/api/sandbox.go
+++ b/internal/api/sandbox.go
@@ -458,7 +458,27 @@ func (s *Server) createSandboxRemote(c echo.Context, ctx context.Context, cfg ty
 	requestedMemoryMB := cfg.MemoryMB
 	requestedCpuCount := cfg.CpuCount
 
+	// Pre-generate sandbox ID so we can create the session in PG before the
+	// gRPC call. The worker's RecordScaleEvent needs the org_id from the
+	// session row, which must exist before the worker looks it up.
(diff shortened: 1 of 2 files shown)


breardon2011 (approved):
approve

PR #179, opened 2026-04-21
Add Sentry error reporting to control plane and worker

motatoes:
## Summary

- New shared `internal/observability` package that initializes the Sentry Go SDK and registers a panic recover. Both are no-ops when `OPENSANDBOX_SENTRY_DSN` is unset, so existing deployments are unaffected until the DSN is set.
- `cmd/server` and `cmd/worker` call `observability.Init(...)` at startup, `defer` the flush, and `defer observability.Recover()` so panics are captured before the process dies.
- Events are tagged with `service` (`control-plane` | `worker`), `region`, `worker_id`, and `mode`. Release is fed from the existing `WorkerVersion` ldflag on the worker and a new `ServerVersion` ldflag on the control plane.
- New config fields: `OPENSANDBOX_SENTRY_DSN`, `OPENSANDBOX_SENTRY_ENVIRONMENT` (defaults to `OPENSANDBOX_REGION`), `OPENSANDBOX_SENTRY_SAMPLE_RATE` (default `1.0`), `OPENSANDBOX_SENTRY_TRACES_SAMPLE_RATE` (default `0.0` — tracing off).
- `deploy/{server,worker}.env.example` updated to document the env vars and the `server-sentry-dsn` / `worker-sentry-dsn` Key Vault secret names.

Out of scope for this PR (deferred): oc CLI and the TS / Python SDKs.

## Test plan

- [ ] `go build ./cmd/server ./cmd/worker ./internal/observability` passes
- [ ] `go vet ./...` clean on touched packages
- [ ] Start control plane locally with `OPENSANDBOX_SENTRY_DSN` unset — log line `sentry: enabled` is absent, no new behavior
- [ ] Start control plane locally with the DSN set — log line appears; trigger a panic and confirm the event shows up in Sentry with `service=control-plane` tag
- [ ] Same two checks for the worker — confirm `service=worker` tag on the event
- [ ] Add `server-sentry-dsn` / `worker-sentry-dsn` to Azure Key Vault before merging to prod

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #180, opened 2026-04-21
Register sentry DSN in key vault secret mapping

motatoes:
## Summary

Follow-up to #179. Without this, `LoadSecretsFromKeyVault` silently ignores `server-sentry-dsn` and `worker-sentry-dsn` (only names in the hardcoded `secretMapping` whitelist are loaded), so `OPENSANDBOX_SENTRY_DSN` stays empty and Sentry init no-ops even after setting the secrets in the vault.

## Test plan

- [ ] Set `server-sentry-dsn` and `worker-sentry-dsn` in the vault (already done)
- [ ] Redeploy control plane — confirm `keyvault: loaded N secrets` log line includes one more than before, and `sentry: enabled (service=control-plane, ...)` appears at startup
- [ ] Redeploy worker — same check, `service=worker`
- [ ] Trigger a test panic (or `sentry.CaptureMessage`) and confirm the event lands in Sentry


breardon2011 (approved):
approve

PR #186, opened 2026-04-23
Migration Stability

breardon2011:
# Migration Stability — root-cause sweep + cross-golden architecture rewrite

 A debugging effort that started as "Igor's clawputer checkpoint can't be forked" and ended up uncovering seven independent bugs in the migration
  / wake / autoscale path, plus a structural rewrite of how checkpoints relate to base images.

  By the end the goal shifted: stop trying to make `qemu-img rebase` work across goldens at all. Pin checkpoints to their original base forever; bake the
  previous golden's `default.ext4` into each new AMI so 99% of cross-golden forks are zero-cost.

  ## Summary

  - **Bugs 1–7**: discrete fixes for golden hash collisions, blob existence checks, migration timeouts, scale-down racing replacements, drain-stuck states,
  agent-upgrade poisoning the wake path, and bogus AMI-build version stamping.
  - **Bug 8**: scheduler used committed memory for migration admission; switched to actual RSS.
  - **Bug 9**: internal savevm snapshots in qcow2 made `qemu-img rebase` fragile; extract memory to external `mem.zst` instead.
  - **Bug 10**: golden version was thrashing (new value on every AMI build, even with no rootfs change). Three fixes land it stable: content-addressed rootfs
  cache, deterministic ext4 UUID, narrowed workflow triggers.
  - **Bug 11**: cross-golden `qemu-img rebase -b` block-copy corrupts the overlay's relationship to the captured memory state. Replaced with **pin-to-base**:
  never rebase, just keep checkpoint pointed at its original golden's `default.ext4`.
  - **Bug 12**: each new AMI now bakes in the previous golden's `default.ext4` so the common cross-golden case (customers on the previous golden when an AMI
  rolls) skips the on-demand blob download.

  ---

  ## Bugs 1–7 (original sweep)

  ### 1. `golden_version` was a 64-bit hash of only the first 1 MB

  `computeGoldenVersion` hashed the first 1 MB of `default.ext4` and truncated to 16 hex chars. Two base images with identical first-MB (common — ext4
  superblock + boot region are often untouched across rebuilds) collided to the same version string despite differing tail bytes. `ensureCheckpointRebased`
  thought no rebase was needed → overlay reads fell through to the new base's differing bytes → `EXT4-fs error: Directory block failed checksum` inside forked
   guests.

  **Fix:** hash the whole base file. ~20 s one-time on worker boot.

  **Impact:** original root cause of the "Igor's clawputer checkpoint is broken across forks" problem that drove weeks of speculation about agent.sock and
  cross-worker failures.

  ### 2. `storage.Exists` treated 404 as an error

  `CheckpointStore.Exists` returned `(false, err)` for any `Head()` failure, and Azure/S3 SDKs return a typed error on missing blobs. `uploadBaseImageIfNew`
  then bailed out without uploading. Old-golden bases were never archived → cross-golden migration couldn't download the old base → drain stuck.

  **Fix:** new `storage.ErrNotFound` sentinel; each backend's `Head` translates its native 404 into it; `Exists` maps `ErrNotFound` to `(false, nil)`.

  ### 3. Migration's "complete" step had a 30 s timeout, source destroyed on timeout

  `scaler.liveMigrateSandbox` called `CompleteMigrationIncoming` with a 30 s context. Post-QMP sequence (agent reconnect, network patch, clock sync,
  virtio-mem sync) can exceed 30 s on a freshly resumed post-rebase VM. Timeout → scaler errors → source's drain proceeds and destroys it → target unfinalized
   → sandbox lost. This destroyed `sb-a7f54cc7` during round 1.

  **Fix:** bumped to 3 min. Post-QMP work took ~4 s in round 2 — 30 s wasn't tight, it was wrong.

  ### 4. Smart scale-down destroyed the rolling-replace target

  On scaler startup, `targetWorkerVersion` is unknown until the first AMI/KV refresh. During the gap, `smartScaleDown` couldn't distinguish a stale worker
  from the newly-launched replacement and would drain the new one (low utilization + 0 sandboxes).

  **Fix:** gate `smartScaleDown` on a known `targetWorkerVersion` when any worker reports a non-empty `WorkerVersion`. Force AMI refresh on the first evaluate
   tick. Round 1 destroyed `f2d35c9a` for this reason; round 2+ did not.

  ### 5. Drain fell into `waitForNaturalDrain` after 3 migration failures

  After 3 consecutive migration failures, `drainWorker` called `waitForNaturalDrain` — a terminal poll loop. For persistent sandboxes (`--timeout 0`) drain
  stuck for the full 45-min `drainTimeout` and required manual Redis + CP restart.

  **Fix:** replace fall-through with a 5-min backoff + reset `migrationFailures = 0` + continue. Transient failures (target booting, base just uploaded, stale
   dir cleared) auto-recover. Permanent failures still terminate on `drainTimeout`, but a fresh drain kicks off on the next eval tick. Bonus:
  `PrepareIncomingMigrationWithS3` now `RemoveAll(sandboxDir)` before writing so zstd doesn't trip on a leftover `workspace.qcow2`.

  ### 6. Wake path pushed a 12 MB agent upgrade while the guest was post-resume-thrashing

  `doWake` synchronously called `upgradeAgentIfNeeded` right after `cont`. Guest is still finishing disk I/O replay + scheduler recovery. `WriteFileBinary`
  streams 12 MB in 512 KB chunks; gRPC keepalive `Time=30s/Timeout=10s` — keepalive ping misses ACK during thrash → transport tears down mid-chunk →
  `keepalive ping failed to receive ACK within timeout`. Abort path leaves `vm.agent` holding a poisoned `grpc.ClientConn`. Every subsequent RPC returns `use
  of closed network connection` — the exact symptom we'd been chasing for days.

  **Fix (three parts):**
  1. Move upgrade off wake hot path (`go m.upgradeAgentIfNeeded(...)`). Wake returns immediately with current-version agent.
  2. `waitAgentStable(vm, 30s)` polls `Ping` with a 500 ms deadline. Three consecutive fast pings = past resume thrash.
  3. On upgrade failure, close the broken `grpc.ClientConn`, `waitForAgentSocket` for a 15 s reconnect, swap in fresh client or set `vm.agent = nil`. No more
  poisoned conns.

  ### 7. Build script hardcoded `AgentVersion` to top-level repo SHA

  Both worker and agent binaries got their version stamped from top-level `HEAD`. Any commit triggering an AMI build (including `round.txt` bumps or unrelated
   PRs) made the worker expect agent version `<new-SHA>` while baked agent reported the same. Fresh sandboxes match. But existing sandboxes (forked
  checkpoints, woken hibernates) report the OLD SHA → mismatch → `upgradeAgentIfNeeded` triggers Bug 6's race — even though `cmd/agent/` is byte-identical.

  **Fix:** scope `AGENT_VERSION` to commits that actually touched agent code:
  ```
  AGENT_VERSION=$(git log -1 --pretty=format:%h -- cmd/agent internal/agent proto/agent)
  [ -z "$AGENT_VERSION" ] && AGENT_VERSION=$VERSION
  ```

  ---

  ## Bug 8 — Scheduler used committed memory for migration admission

  `scaler.findMigrationTarget` and worker capacity checks used the sandbox's **committed** memory (the QEMU `-m maxmem` value, often 16 GB) to decide if a
  migration target had room. A 16 GB sandbox actually using 800 MB RSS was being rejected from any worker with less than 16 GB of free committed budget, even
  when 14 GB of physical RAM was idle.

  **Fix:** new `VMActualMemoryMB(sandboxID)` reads `/proc/[pid]/status` `VmRSS`. `PreCopyDrivesResponse` carries the source's actual memory in a new
  `actual_memory_mb` field. `findMigrationTarget` and worker admission both use it. `Drain` triggers scale-up when no viable target exists. Hardcoded
  `TargetMemoryMb: 4096` removed.

  ---

  ## Bug 9 — Internal savevm snapshots in qcow2 broke `qemu-img rebase`

  Checkpoints were created with `savevm` → memory dump becomes an internal qcow2 snapshot. `qemu-img rebase` across goldens has known problems with internal
  snapshots: rebase to the new base, internal snapshot's memory state references old-base disk content, guest crashes on first dirty page flush.

  **Fix:** new `extractCheckpointMemory` (`internal/qemu/checkpoint_memextract.go`) spawns a throwaway QEMU with `-S -loadvm`, migrates state to an external
  `mem.zst` via `migrate exec:zstd`, then strips the internal snapshot. Forks load with `-incoming exec:zstdcat mem.zst`. qcow2 is now clean (no internal
  snapshots) so any rebase-related operation is simpler.

  ---

  ## Bug 10 — Golden version thrashed on every AMI build

  Each Packer build produced a different `default.ext4` even when no rootfs input had changed: `mkfs.ext4` stamps a random UUID + hash seed → different bytes
  → different SHA256 → new goldenVersion → every existing checkpoint becomes "cross-golden" → forks pay rebase cost (or break).

  **Three fixes, all required:**

  1. **Content-addressed rootfs cache.** Packer hashes the agent binary + `deploy/firecracker/rootfs/**` + scripts + guest kernel modules. On cache hit,
  downloads the existing `default.ext4` from `rootfs-cache/{hash-short}/default.ext4` instead of rebuilding. Same inputs → same cached artifact → same
  goldenVersion across AMIs.
  2. **Deterministic ext4.** `build-rootfs-docker.sh` accepts `ROOTFS_UUID` env var. Packer derives UUID from the input hash and passes it. `mkfs.ext4 -U
  $UUID -E hash_seed=$UUID` makes the output byte-stable, so the file hash matches across workers even on cache races.
  3. **Narrow workflow triggers.** `build-worker-ami.yml` paths reduced to inputs that actually matter for the worker AMI; `internal/api/**`,
  `internal/billing/**`, web, docs, markdown all excluded via `!` negations (GitHub forbids combining `paths` + `paths-ignore`).

  After this, `cmd/server/**` or web changes don't bump golden. Round 8 verified: a no-change build kept goldenVersion stable.

  ---

  ## Bug 11 — Cross-golden `qemu-img rebase -b` corrupts checkpoints (Pin-to-base)

  Even with the extract-memory format and stable golden hashing, **cross-golden rebase fundamentally cannot work**: the captured memory state holds VFS
  caches, dentries, and dirty pages that reference the OLD base's disk content. Swapping in a new base under it tears reads the moment the kernel flushes a
  dirty page. We empirically validated this in three "Variant" attempts (in-flight rebase + same-host live-migrate + flatten-then-rebase) before concluding
  it's structural, not a fix-the-call-site problem.

  **Architectural fix: never rebase across goldens.** Checkpoints are now permanently tied to the goldenVersion they were created against.

  `ensureCheckpointRebased` (kept the name for call-site stability) now:
  1. Reads the checkpoint's pinned `goldenVersion` from `snapshot-meta.json`.
  2. Resolves a local path: current `default.ext4` if versions match → `bases/{ver}/default.ext4` if retained on disk → on-demand blob download (cached
  persistently).
  3. `qemu-img rebase -u` — metadata-only repoint of the overlay's `backing_file` field. ~1 ms. No block movement, no internal-snapshot corruption surface.
  4. `meta.GoldenVersion` is **never mutated**.

  Also removed: `rebaseCheckpointToCurrentBase`, `migrateStaleCheckpoints` background sweep, `backfillGoldenVersions`, `reuploadCheckpoint`,
  `acquireOldBase`'s `/tmp` refcount cleanup. All replaced by lazy on-demand resolve. `migration.go`'s overlay-mode rebase now uses the same resolve + `rebase
   -u` flow.

  ---

  ## Bug 12 — Previous-golden retention in AMI

  With pin-to-base, cross-golden forks pull the old base from blob on first use. ~4 GB download takes ~6.5 s on Azure with parallel chunking — fine, but
  unnecessary in the common case where customers are pinned to the **previous** AMI's golden when a new one rolls.

  **Packer changes:** new `prev_golden_version` var. New provisioner step downloads `bases/{prev}/default.ext4` from blob and bakes to
  `/opt/opensandbox/images/bases/{prev}/default.ext4`. Non-fatal on download failure (rollout 1 has nothing to retain).

  **Workflow changes:** before Packer runs, read `golden-version` from Key Vault → pass to Packer. After successful build, extract new goldenVersion from
  packer output and write back to KV. Two rollouts to bootstrap; steady-state every AMI ships with current + 1 previous.

  **Worker first-boot:** Azure custom-data copies `/opt/opensandbox/images/bases/*` → `/data/firecracker/images/bases/*` so retained bases live on NVMe where
  QEMU expects them.

  **Validation in production (rollouts 1–3):**
  - `/opt/opensandbox/images/bases/1145375f99bdc717/default.ext4` baked into AMI 1.0.35 (4 GiB, mtime = Packer build).
  - Fresh worker on 1.0.35 logged `cloud-init: Copied retained bases from /opt/...` at boot.
  - Worker journal showed **zero** `downloading base 1145375f99bdc717` entries — AMI-baked copy was sufficient.
  - Separate worker without retention (rolled before CP update) downloaded `dc211d8b55f71760` for an older checkpoint in 6.5 s and cached it correctly.
  Pin-to-base on-demand path verified end-to-end.

  ---

  ## Test plan

  - [x] `go test ./internal/qemu/` — unit tests for resolve / retain / download paths, concurrent-download dedup, legacy-checkpoint fallback
  - [x] `go build ./...` — full repo compiles on linux/amd64
  - [x] Three production AMI rollouts on `deployment-test`, validated:
    - [x] Rollout 1 seeds KV with new `golden-version`
    - [x] Rollout 3 retention step downloads + bakes prev base into AMI
    - [x] Fresh worker spawns with retained base on NVMe at boot, no blob redownload
    - [x] Cross-golden checkpoint fork on rolled worker resolves base from blob, caches on NVMe, succeeds without `rebase -b` block copy
  - [ ] Smoke: fork an existing prod customer checkpoint after merge to confirm pin-to-base behaves identically in `opencomputer-prod` (will run after merge)


.claude/scheduled_tasks.lock
```diff
@@ -0,0 +1 @@
+{"sessionId":"d019cd9c-1713-4439-8a80-4efe09b5ebda","pid":47199,"acquiredAt":1776981820229}
```
motatoes: remove this file ?

motatoes (approved):

PR #188, opened 2026-04-24
exec+sdk: stateful shell sessions with reattach

motatoes:
## Summary

- New `exec.shell()` on both TS + Python SDKs: a long-lived bash exec session with a stable `sessionId`, where `cwd`, exported env vars, and shell functions persist across `.run()` calls. Terminal-tab ergonomics, single server-side process.
- New `exec.reattachShell(sessionId)` / `exec.reattach_shell(session_id)`: revisits an open shell from a different client (or a different process invocation). No backend changes required — the worker already sends `0x04` scrollback-end markers and exec sessions survive client disconnect.
- Per-`run()` framing uses dual sentinels: a DONE marker on stdout and an EXIT marker on stderr. A run resolves only once **both** markers are observed. Without this, the stderr sentinel frequently lands before the user's stdout is drained (stdout/stderr are read on separate goroutines in the agent) and the trailing stdout is silently dropped into the next run.
- `ShellImpl` gates on scrollback completion: inbound chunks before the `0x04` marker are ignored, so reattaching to a shell that has prior runs in its scrollback doesn't mis-parse old sentinels as a current completion.

## Incidental fixes found while wiring this up

- TS `Exec.attach()` now awaits `ws.onopen` before returning. The previous code returned while the WS was still CONNECTING, and the first `sendStdin()` was silently dropped by the `readyState !== OPEN` guard in `sendStdin`.
- TS: fixed a bash syntax error in the command wrapper (`{ cmd\n; }` → `{ cmd\n}`). The stray `;` after the newline was producing `bash: syntax error near unexpected token ';'` and bash exited with code 2 on the first run.
- Python `Sandbox.exec` now pairs URL + credential correctly. Previously the property passed the sandbox JWT to whichever URL was available; when falling back to `_api_url` (control plane), the JWT was rejected with an `InvalidStatus` handshake. Control-plane requests now use the API key and include the `/api` prefix.

## Terminal-tab semantics

`exit N` inside a `run()` kills the shell — same as closing a terminal tab — and rejects the pending run with `ShellClosedError`. Tests updated to match (step 5 originally asserted `exit 42 || true; ( exit 7 )` returned 7; now uses a subshell so the outer shell stays alive).

## Test plan

- [ ] `npx tsx sdks/typescript/examples/test-shell.ts` (hits `app.opencomputer.dev`): all assertions pass through step 15, including reattach preserving cwd/env and `exit N` closing the shell.
- [ ] `python sdks/python/examples/test_shell.py`: same assertions, Python side.
- [ ] `sandbox.exec.list()` still sees the shell session alongside regular exec sessions; `shell.sessionId` is the same ID.
- [ ] Existing `exec.run()` / `exec.start()` / `exec.background()` behavior unchanged.

### DIFF
--- a/docs/reference/python-sdk.mdx
+++ b/docs/reference/python-sdk.mdx
@@ -294,9 +294,9 @@ Run a command synchronously via `sh -c`.
 result = await sandbox.exec.run("npm test", cwd="/app")
 ```
 
-### `await sandbox.exec.start(command, ...): str`
+### `await sandbox.exec.start(command, ...): ExecSession`
 
-Start a long-running command. Returns the **session ID** (not a session object).
+Start a long-running command with streaming I/O. Returns an `ExecSession` — see the [full reference](/reference/python-sdk/exec#await-sandbox-exec-start-command).
 
 | Parameter | Type | Default | Description |
 | --- | --- | --- | --- |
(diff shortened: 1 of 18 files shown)


breardon2011 (approved):
Approve

PR #189, opened 2026-04-24
oc exec: stream by default, implement `oc exec attach`

motatoes:
## Summary

Fix two CLI UX holes:

1. **`oc exec <sb> -- cmd`** (default, no flags) now creates a session, **attaches over WebSocket, and streams stdout/stderr live** to the user's terminal while forwarding stdin. The CLI exits with the command's exit code. Previously this path just POSTed to `/exec` and printed "Session X created / attach with ..." — a dead-end for anyone who actually wanted output.

2. **`oc exec attach <sb> <sid>`** was a stub that printed "not yet implemented in CLI — use the SDK or websocat." It now works: replays scrollback, streams live, forwards stdin, exits with the process exit code.

Both use the same binary WS protocol the SDKs speak (`0x00` stdin, `0x01` stdout, `0x02` stderr, `0x03` exit (big-endian int32), `0x04` scrollback-end). No backend changes — all the server-side plumbing for this has existed since exec sessions shipped.

## Flag changes

| Flag | Behavior |
| --- | --- |
| *(no flag)* | **New default:** create session + stream + exit with process exit code |
| `--detach` | **New:** create session, print id + hint, don't wait (preserves old no-flag behavior for fire-and-forget scripts) |
| `--wait` | Unchanged: buffered `/exec/run` POST, no streaming, prints result at the end (mutually exclusive with `--detach`) |

## Ctrl-C handling

First Ctrl-C sends SIGINT to the remote process via `/exec/:sid/kill` and keeps streaming — the process decides whether to handle it. Second Ctrl-C force-closes the WS and exits the CLI locally (the remote process keeps running; use `oc exec kill` to terminate it).

## Test plan

- [ ] `oc exec <sb> -- echo hello` — streams \"hello\", exits 0
- [ ] `oc exec <sb> -- sh -c 'echo out; echo err >&2; exit 7'` — stdout and stderr visibly separated, CLI exits 7
- [ ] `oc exec <sb> -- sleep 5` with Ctrl-C — first Ctrl-C interrupts, prints `^C sent SIGINT`; CLI exits when sleep is killed
- [ ] `oc exec <sb> -- cat` with piped stdin (`echo foo | oc exec ...`) — `foo` echoed back
- [ ] `oc exec <sb> --detach -- long-job` — prints session id, exits immediately
- [ ] `oc exec attach <sb> <sid>` against a detached session — scrollback replayed, live streaming continues
- [ ] `oc exec <sb> --wait -- cmd` — unchanged behavior, buffered result
- [ ] `oc exec <sb> --wait --detach -- cmd` — rejected with clear error

## Follow-ups (not in this PR)

- Fix typical Ctrl-C-from-pty issue: when stdin isn't a tty we rely on OS signal delivery; terminal raw-mode for richer key forwarding could come later.
- Consider making `--wait` delegate to the streaming path too (streaming is a strict superset). Not done here to avoid breaking scripts that parse the final buffered output.

### DIFF
--- a/cmd/oc/internal/commands/exec.go
+++ b/cmd/oc/internal/commands/exec.go
@@ -1,13 +1,27 @@
 package commands
 
 import (
+	"context"
+	"encoding/binary"
 	"fmt"
 	"os"
+	"os/signal"
 	"strings"
+	"syscall"
 
+	"github.com/gorilla/websocket"
 	"github.com/opensandbox/opensandbox/cmd/oc/internal/client"
 	"github.com/opensandbox/opensandbox/pkg/types"
 	"github.com/spf13/cobra"
+)
+
+// Exec WebSocket binary protocol stream IDs.
+const (
+	execStreamStdin         = 0x00
+	execStreamStdout        = 0x01
+	execStreamStderr        = 0x02
+	execStreamExit          = 0x03
+	execStreamScrollbackEnd = 0x04
 )
 
 var execCmd = &cobra.Command{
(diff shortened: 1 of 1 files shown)


breardon2011 (approved):
approve

PR #190, opened 2026-04-24
oc: add `oc update` + update-available nag

motatoes:
## Summary

Two CLI self-maintenance pieces.

### 1. `oc update`

Fetches the latest release from GitHub (`diggerhq/opencomputer`), picks the asset matching `runtime.GOOS/GOARCH` (`oc-linux-amd64`, `oc-darwin-arm64`, etc.), verifies its SHA256 against the release's `checksums.txt`, and atomically replaces the running binary via `rename(2)`.

Flags:
- `--check` — report whether an update is available, don't install
- `--yes` — skip the confirmation prompt

Behavior:
- Refuses to self-update a dev build (`Version == "dev"`) — you'd overwrite your local dev binary with a published release, which is rarely what you want.
- Clear error when the binary path isn't user-writable (\"no write access to /usr/local/bin (re-run with sudo …)\"); otherwise it would fail cryptically at rename time.
- Uses `os.WriteFile` to a sibling `.new` path + `os.Rename`, so there's no window where the binary is half-written.

### 2. Background update-check

New `PersistentPostRun` hook that prints a one-line nag to stderr when a newer release is available:

```
A new oc release is available: v0.5.0.72 (current: v0.5.0.67). Run \`oc update\` to install. Disable this check with OC_NO_UPDATE_CHECK=1.
```

Rules:
- Checks at most once per 24h. Result cached in `$XDG_CACHE_HOME/opencomputer/update-check.json`.
- 2s network timeout — a slow or blocked GitHub API never slows down the CLI.
- Silent on dev builds, non-TTY stderr (so CI/pipelines stay clean), and when `OC_NO_UPDATE_CHECK=1`.
- Skipped for `oc update`, `oc help`, and shell-completion scaffolding.
- Any error (network, parse, cache read/write) is swallowed — version checking must never surface failures to the user.

## No server changes

The release pipeline (`.github/workflows/release-cli.yml`) already publishes `oc-<os>-<arch>` binaries + `checksums.txt` on every push to main. This PR is a pure client against that existing surface.

## Test plan

- [ ] `oc update --check` on a dev build → prints current/latest, \"dev build — skipping update\"
- [ ] `oc update --check` on a build older than latest → prints \"update available: run \`oc update\`...\"
- [ ] `oc update --check` on a build equal to latest → prints \"already up to date\"
- [ ] `oc update --yes` on an older build → downloads, verifies SHA, replaces; `oc --version` reports the new version
- [ ] `oc update` without `--yes` → prompts; empty input or \"y\" proceeds, anything else aborts
- [ ] `oc update` on a read-only binary (e.g. system /usr/local/bin without sudo) → clear permission error, no partial write
- [ ] `oc sandbox ls` in a real terminal with an older version → nag line appears on stderr once
- [ ] `oc sandbox ls 2>/dev/null` → nag suppressed (non-TTY stderr)
- [ ] `OC_NO_UPDATE_CHECK=1 oc sandbox ls` → nag suppressed
- [ ] Second `oc sandbox ls` within 24h → uses cached check (no network call)
- [ ] Run with network disabled → nag silent, no error surfaced

### DIFF
--- a/cmd/oc/internal/commands/root.go
+++ b/cmd/oc/internal/commands/root.go
@@ -39,6 +39,17 @@ var rootCmd = &cobra.Command{
 		printer = output.New(jsonOutput)
 		return nil
 	},
+	PersistentPostRun: func(cmd *cobra.Command, args []string) {
+		// One-line nag to stderr when a newer release is available. Skip
+		// for `oc update` itself, help, and completion scaffolding — see
+		// maybePromptUpdate for the full skip list (dev build, non-TTY,
+		// OC_NO_UPDATE_CHECK).
+		switch cmd.Name() {
+		case "update", "help", "completion", "__complete":
+			return
+		}
+		maybePromptUpdate()
+	},
 	SilenceUsage:  true,
 	SilenceErrors: true,
 }
--- a/cmd/oc/internal/commands/update.go
+++ b/cmd/oc/internal/commands/update.go
@@ -1,0 +1,258 @@
+package commands
+
+import (
+	"bufio"
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/spf13/cobra"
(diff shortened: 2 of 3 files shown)


breardon2011 (approved):
approve

PR #191, opened 2026-04-25
dashboard+skill: first-run onboarding + opencomputer skill rewrite

motatoes:
## Summary
- **Dashboard onboarding** — when an account has zero sessions, the dashboard renders a two-step "Getting Started" panel instead of empty placeholders: (1) the `npx skills add diggerhq/opencomputer` install command with a Copy button, and (2) a "Generate API Key" button that calls the existing `createAPIKey('Default')` mutation and reveals the key + a ready-to-paste `oc config set api-key <key>` command. After the first sandbox runs, the normal dashboard view returns automatically.
- **Skill consolidation** — removes the stale project-local `.claude/skills/openqemu/` skill (raw curl against the OpenSandbox API) in favor of the canonical `skills/opencomputer/` skill (uses the `oc` CLI).
- **Skill onboarding rewrite** — the opencomputer skill now has an explicit "First-time setup" section that runs before any other `oc` command: it checks for `oc` and installs via the official `scripts/install.sh` one-liner if missing, then checks `oc config show` and (on missing key / 401) opens https://app.opencomputer.dev in the user's browser and instructs them to paste `oc config set api-key …` into their own terminal. Allowed-tools widened so install + browser-launch don't trigger permission prompts.

## Test plan
- [ ] `make infra-up && make run-pg-workos` + `make web-dev`, sign up with a fresh org, confirm the Dashboard shows "Welcome to OpenComputer" with both step cards.
- [ ] Click "Generate API Key" → key reveal banner appears with both Copy buttons working; subsequent reload shows "You already have 1 API key" with a working "Manage keys" link to `/api-keys` (no redirect loop).
- [ ] Create a sandbox → reloading the Dashboard now shows the regular stat cards / Live Sandboxes / Recent Sessions view.
- [ ] In a fresh shell (no `oc` installed, no API key configured), invoke the opencomputer skill and confirm it (a) installs the CLI, (b) opens the dashboard in the browser, and (c) waits for the user before running sandbox commands.


breardon2011 (approved):
approve

PR #219, opened 2026-05-04
fix(sandbox): default networkEnabled=true so dashboard stops mislabeling sandboxes

motatoes:
## Summary

`SandboxConfig.NetworkEnabled` was a non-pointer `bool`, so any client that omitted the field — `oc sandbox create`, `oc checkpoint spawn`, the TS SDK, `sessions-api`'s agent-create path — ended up with the zero value `false`. That value got persisted into `sandbox_sessions.config_json` and rendered in the dashboard as **"Disabled"** even though the worker (`internal/qemu/manager.go` `Create`) sets up the TAP/DNAT unconditionally and the sandbox actually has networking. Forks inherited the same `false` from the parent checkpoint's persisted config, so spawning from a checkpoint produced more "Disabled" sandboxes.

The controlplane already had a `*bool` + nil-default-to-true path (`internal/controlplane/server.go:112-115`), but the comment there gave the game away: "the worker currently sets up networking unconditionally, so a missing field that marshaled to false caused the UI to mislabel sandboxes as 'Disabled'." The API server and fork-from-checkpoint flow didn't have the same defaulting, so the fix never reached the sandboxes most users actually create.

## Changes

- **`pkg/types/sandbox.go`** — flip `NetworkEnabled bool` → `*bool`; add `IsNetworkEnabled()` (nil → true) and `EnsureNetworkEnabledDefault()` helpers so the rule lives in one place.
- **`internal/api/sandbox.go`**
  - `createSandbox`: call `cfg.EnsureNetworkEnabledDefault()` after `Bind`, so the value persisted to `sandbox_sessions.config_json` is explicitly `true` when the client omits the field.
  - `createFromCheckpointCore`: default `originalCfg` after unmarshal so forks of older checkpoints (which were persisted with `false` from the previous bug, or with no value at all) come up as Enabled.
  - Both gRPC `CreateSandboxRequest` sites use `IsNetworkEnabled()`.
- **`internal/api/image_builder.go`** — combined-mode build path now assigns a `*bool` instead of the bool literal.
- **`internal/worker/grpc_server.go`** — wraps the proto `bool` into `*bool` when constructing the internal `SandboxConfig`.

## Test plan

- [x] `go build ./pkg/types ./internal/api ./internal/controlplane ./internal/worker ./internal/qemu ./internal/sandbox ./cmd/oc/...` — clean.
- [x] `go test ./pkg/types/... ./internal/api/...` — pass.
- [ ] Create a sandbox via `oc sandbox create` and confirm the dashboard now shows **Enabled**.
- [ ] Fork a checkpoint via `oc checkpoint spawn` and confirm the resulting sandbox also shows **Enabled** (including for forks of pre-fix parents).
- [ ] Confirm explicit `{"networkEnabled": false}` over the raw API still persists as Disabled (the pointer correctly distinguishes unset vs explicit false).

### DIFF
--- a/internal/api/image_builder.go
+++ b/internal/api/image_builder.go
@@ -399,7 +399,8 @@ func (s *Server) buildImage(ctx context.Context, orgID uuid.UUID, manifest *Imag
 		}
 		_ = resp
 	} else if s.manager != nil {
-		cfg.NetworkEnabled = true
+		t := true
+		cfg.NetworkEnabled = &t
 		sb, err := s.manager.Create(ctx, cfg)
 		if err != nil {
 			return uuid.Nil, fmt.Errorf("failed to create build sandbox: %w", err)
--- a/internal/api/sandbox.go
+++ b/internal/api/sandbox.go
@@ -26,6 +26,9 @@ func (s *Server) createSandbox(c echo.Context) error {
 			"error": "invalid request body: " + err.Error(),
 		})
 	}
+	// Default networkEnabled=true when caller omits it, so the value persisted
+	// to sandbox_sessions.config_json is explicit and forks inherit it correctly.
+	cfg.EnsureNetworkEnabledDefault()
 
 	// Validate CPU/memory against allowed tiers.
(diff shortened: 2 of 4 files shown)


breardon2011 (approved):
Approved

PR #221, opened 2026-05-04
fix(secrets-proxy): register session for allowlist-only secret stores

motatoes:
## Summary
- Register a proxy session whenever a sandbox has tokens **or** an egress allowlist; previously the session was only created when the store had ≥1 sealed secret, so allowlist-only stores produced `407 no_session` for every outbound request and the allowlist was dead config.
- Mirror the same fix on the wake / restore-from-checkpoint / cross-worker-migration re-register paths so hibernated sandboxes recover allowlist enforcement on resume.
- Stop `qemu.sealSandboxEnvs` from early-returning when both env maps are empty if an allowlist is set — otherwise `HTTP_PROXY` is never injected and the sandbox bypasses the proxy entirely (worse failure mode: unrestricted egress).

## Background
Reported by USER sandbox couldn't reach an ngrok HOST even though it was in their `XXXX` store's egress allowlist. Worker logs showed the proxy receiving CONNECTs and rejecting with `audit src=… action=rejected reason=no_session`. Root-caused to `internal/secretsproxy/proxy.go` registering a session only when `len(tokenMap) > 0`. Their store had 12 hosts in the allowlist but zero secret entries.

Same `> 0` guard existed in three lifecycle paths:
- `internal/qemu/manager.go:2713` (RestoreFromCheckpoint)
- `internal/qemu/snapshot.go:490` (wake from hibernation)
- `internal/qemu/migration.go:751` (cross-worker migration)

All four are updated to also register/re-register when there's an allowlist alone.

## Test plan
- [x] New unit tests in `internal/secretsproxy/proxy_test.go` covering: allowlist-only registers session, allowlist-only enforces correctly, no-config returns nil, secrets-only still registers (regression).
- [x] `go test ./internal/secretsproxy/...` passes.
- [x] `GOOS=linux GOARCH=amd64 go build ./internal/qemu/...` succeeds.
- [ ] Manual e2e: create allowlist-only store via `oc secret-store create --egress example.com`, create sandbox with at least one env var, `curl https://example.com/` from inside → expect HTTP 200 (was 407 before fix). `curl https://evil.com/` → expect 403 host-not-in-allowlist.
- [ ] Manual e2e: hibernate then wake an allowlist-only sandbox; verify allowlist still enforces.

## Out of scope (separate work)
- Live propagation of `UpdateSecretStore` to running sandboxes (today running sandboxes still need recreation to pick up store updates — same limitation as before this PR).
- Fork-path missing `RecordScaleEvent` billing leak — separate PR.

## Workaround for affected users (no code change)
Add any secret entry to the store, then recreate the sandbox:
```
oc secret set <store-id> KEEPALIVE x
oc sandbox kill <sandbox-id>   # recreate via your normal flow
```

### DIFF
--- a/internal/qemu/manager.go
+++ b/internal/qemu/manager.go
@@ -300,7 +300,7 @@ func (m *Manager) sealSandboxEnvs(ctx context.Context, sandboxID string, netCfg 
 		}
 		return merged
 	}
-	if len(cfg.Envs) == 0 && len(cfg.SecretEnvs) == 0 {
+	if len(cfg.Envs) == 0 && len(cfg.SecretEnvs) == 0 && len(cfg.EgressAllowlist) == 0 {
 		return cfg.Envs
 	}
 	sealed := m.secretsProxy.CreateSealedEnvs(sandboxID, netCfg.GuestIP, netCfg.HostIP, cfg.Envs, cfg.SecretEnvs, cfg.EgressAllowlist, cfg.SecretAllowedHosts)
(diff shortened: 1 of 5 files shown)


breardon2011 (approved):
approve

PR #222, opened 2026-05-04
fix(billing): record scale event on fork-from-checkpoint

motatoes:
## Summary
- Forks-from-checkpoint never produced a `sandbox_scale_events` row, so the usage-reporter excluded the org from `ListFreeOrgIDsWithOpenUsage` / `ListBillableOrgIDs` and no credits or Stripe usage were reported for fork-derived runtime. Both tiers leaked.
- Two complementary fixes are needed:
  - **worker:** `CreateSandbox` returned early on the warm-fork paths (local-cache + S3-fallback) before reaching the existing `RecordScaleEvent` block. Extract `recordInitialScaleEvent` and call from all three success-return paths.
  - **CP:** `createFromCheckpointCore` wrote the `sandbox_sessions` row *after* the worker gRPC call returned. Even with the worker fix, `GetSandboxOrgID` would resolve to empty and the helper would no-op. Pre-write the row with `status='pending'` before the gRPC call, transition to `running` / `failed` after — mirrors the existing from-scratch pattern.
- Adds `tests/manual/verify-billing-and-proxy-fixes.sh` — end-to-end verification covering this fix and the related allowlist-only proxy fix from #221. New `tests/manual/` directory for live-env scripts that can't run from `go test`.

## Why both changes
Either fix alone leaves the bug intact:
| State | Behavior on fork |
|---|---|
| CP pre-write only | Worker still returns before any `RecordScaleEvent` call → no row |
| Worker fix only | `GetSandboxOrgID` returns empty (no row yet) → helper no-ops → no row |
| Both | Helper runs, lookup returns the org, row inserted |

An alternative would be passing `org_id` in `CreateSandboxRequest` to drop the DB lookup, but mirroring the from-scratch CP pattern keeps both create paths consistent and avoids a proto change.

## Test plan
- [x] `GOOS=linux GOARCH=amd64 go build ./internal/api/... ./cmd/server ./cmd/worker` succeeds.
- [x] End-to-end verified on dev: `tests/manual/verify-billing-and-proxy-fixes.sh` passes — fork produces a `sandbox_scale_events` row with the expected memory/cpu/disk values.
- [ ] After merge, monitor `usage-reporter` logs for healthy free-tier deductions on accounts that previously had `last_usage_reported_at` frozen due to fork-only activity.

## Out of scope
- Backfill for orgs whose historical fork-only runtime was never billed — separate one-off script.
- Allowlist-only proxy session fix — separate PR (#221).
- Live propagation of `UpdateSecretStore` to running sandboxes.

### DIFF
--- a/internal/api/sandbox.go
+++ b/internal/api/sandbox.go
@@ -2379,6 +2379,30 @@ func (s *Server) createFromCheckpointCore(c echo.Context, userEnvs map[string]st
 		}
 	}
 
+	// Pre-write the sandbox_sessions row with status='pending' so the worker
+	// can resolve org_id during CreateSandbox — the worker's
+	// recordInitialScaleEvent looks up sandbox→org via this row. Without the
+	// pre-write, fork-path scale events are silently skipped and both free-tier
+	// credit deduction and pro-tier Stripe metering miss the usage. Mirrors
+	// the from-scratch path's CreateSandboxSessionWithStatus(..., "pending")
+	// call. Status flips to 'running' on success / 'failed' on error below.
+	template := originalCfg.Template
+	if template == "" {
+		template = "default"
+	}
+	mergedMeta := map[string]string{}
+	for k, v := range originalCfg.Metadata {
+		mergedMeta[k] = v
+	}
+	for k, v := range userMetadata {
+		mergedMeta[k] = v
+	}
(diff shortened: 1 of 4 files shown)


breardon2011 (approved):
Approved

PR #225, opened 2026-05-05
design: sandbox session logs (capture, ship, surface)

ZIJ:
## Summary

Design doc for giving every sandbox a "Logs" tab that surfaces everything happening inside it — `/var/log/*` plus all platform-exec'd stdout/stderr — with live tail and search.

Architecture: in-guest forwarder embedded in `osb-agent` ships events directly to **Axiom** over HTTPS. Long-lived dataset-scoped **ingest-only** token baked into sandbox env at create time. Control plane is **off the hot path** — only mints the token at sandbox-create and proxies SSE read queries via a new `/api/sandboxes/:id/logs` endpoint.

Closes the gap with E2B's dashboard (which today shows only platform-side process lifecycle — no command text, no stdout, no network).

## Discussion points

Doc is detailed enough to drive implementation but several decisions are deliberately deferred (cost, retention, per-org opt-out, rate caps). See "Open design questions" and "Decisions needed before phase 0" sections.

Draft PR — comments welcome before we cut a working doc / implementation plan.

## Test plan

- [ ] Reviewers agree on Axiom as the backend (vs. self-hosted Loki / S3+Athena)
- [ ] Reviewers agree on shipping from within osb-agent (vs. separate `osb-log-shipper` binary)
- [ ] Reviewers agree on long-lived ingest-only token (vs. per-sandbox rotation)
- [ ] Spoofing-isolation defence-in-depth (`_event_source` validation) is judged sufficient
- [ ] Phasing matches what we want to ship vs. follow-up

### DIFF
--- a/.agents/design/sandbox-session-logs.md
+++ b/.agents/design/sandbox-session-logs.md
@@ -1,0 +1,493 @@
+# Sandbox session logs: capture, ship, and surface
+
+## TL;DR
+
+Give every sandbox a "Logs" tab in the dashboard that shows
+**everything happening inside it** — every line written to
+`/var/log/*`, every byte of stdout/stderr from every command exec'd
+through the platform — with live tail and free-text search.
+
+The shape: a small in-guest log forwarder embedded in `osb-agent`
+ships events directly to **Axiom** over HTTPS, using a long-lived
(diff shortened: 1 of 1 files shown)


ZIJ:
Recreated as #226 with PII removed from the doc. See #226 for the current draft.

PR #240, opened 2026-05-13
Ship platform logs to Axiom via Vector

motatoes:
## Summary

Adds a structured logging pipeline that ships control plane + worker logs to Axiom with a stable host envelope and a `request_id` that joins across services — **including the deploy wiring**, so a single merge of this PR ends with logs flowing in prod (provided the two KV secrets in the runbook below are set).

- **`internal/obslog`** — JSON `slog` package. Bakes host envelope (`service`, `service_id`, `cell_id`, `region`, `hostname`, `host_ip`, `version`) into every log line; per-request fields (`request_id`, `sandbox_id`, `worker_id`) flow via `context.Context`. Also redirects stdlib `log.Printf` through slog so existing call sites emit JSON automatically — no rip-and-replace.
- **Echo middleware** — Replaces `middleware.Logger()` in both servers. Pulls `X-Request-Id` (from `middleware.RequestID()`) and the `:id` URL param into context; emits one structured access-log line per request.
- **Cross-service join** — `internal/proxy/sandbox_api_proxy.go` Director forwards `X-Request-Id` to the worker, so proxied requests appear on both control-plane and worker logs with the same id.
- **`internal/config`** — adds `OPENCOMPUTER_CELL_ID` (default `<region>-default`) and `OPENCOMPUTER_HOST_IP` (auto-detected, link-local filtered). New fields use the product-named prefix; existing `OPENSANDBOX_*` fields untouched.
- **`deploy/vector/`** — Vector configs reading journald, parsing JSON, enriching non-JSON lines with the host envelope, and shipping to a **new** Axiom dataset (separate from the customer `oc-sandbox-logs` dataset). Roles `{worker, control-plane, dev-host}`.
- **`deploy/vector/populate-vector-env.{sh,service}`** — systemd oneshot that runs `Before=vector.service`, fetches **both** `shared-axiom-platform-ingest-token` and `shared-axiom-platform-dataset` from Azure Key Vault via the VM's managed identity (IMDS → AAD → KV REST), writes `/etc/opensandbox/vector.env`. Tokens never live in an image, workflow artifact, or env file checked into git. Failures (no managed identity, no vault, missing secret) exit 0 so a logging credential problem doesn't break the worker boot.
- **`deploy/packer/worker-ami.pkr.hcl`** — Vector + the populator are baked into every new worker AMI. AMI captures with services enabled-but-not-started; first boot wakes them up.
- **`.github/workflows/build-worker-ami.yml`** — pre-tars `deploy/vector/` so Packer's file provisioner has a stable artifact path.
- **`.github/workflows/deploy-server.yml`** — bundles `deploy/vector/` alongside the server binary, uploads to blob storage, runs `install.sh control-plane` on each CP via `az vm run-command`. Idempotent; refreshes config on every deploy.
- **`deploy/azure/deploy-azure-dev.sh`** — rsync excludes extended so secret-bearing local state files don't end up on the VM.

## Environment variables — pre-merge checklist

**Code (binaries on the host)**: nothing strictly required. The new fields have safe defaults:

| Var | Where | Required? | Default | Notes |
| --- | --- | --- | --- | --- |
| `OPENCOMPUTER_CELL_ID` | server + worker `*.env` | optional | `<region>-default` | Set explicitly when a region has multiple cells/pools. New prefix (deliberate — see config.go comment). |
| `OPENCOMPUTER_HOST_IP` | server + worker `*.env` | optional | auto-detected (skips loopback + link-local) | Override in cloud-init if auto-detect picks the wrong interface. |

**Infra (Vector → Axiom)**: `populate-vector-env.service` fetches **both** the token and the dataset name from Key Vault at every boot. No defaults — a missing secret fails Vector's healthcheck loudly rather than silently shipping to a presumed default. Two KV secrets to provision:

| KV secret | Required? | Maps to env var | Notes |
| --- | --- | --- | --- |
| `shared-axiom-platform-ingest-token` | **REQUIRED** | `AXIOM_PLATFORM_TOKEN` | Axiom **ingest** token, capability `Ingest` only, scoped to one dataset only (your platform dataset — not "All datasets"). Do not reuse the customer-side `shared-axiom-ingest-token`. |
| `shared-axiom-platform-dataset` | **REQUIRED** | `AXIOM_PLATFORM_DATASET` | The Axiom dataset name (e.g. `oc-platform-logs`, `oc-platform-logs-staging`). Kept in KV (not in code) so prod / staging / dev can have separate streams without code changes. |

The VM's managed identity (`AzureWorkerIdentityID` for workers; equivalent on the CP) needs `Get` on both — same RBAC pattern as the existing shared secrets.

**Axiom prerequisites** (do this in Axiom UI before merging):
1. Create your platform-logs dataset (e.g. `oc-platform-logs`).
2. Create an Axiom API token: capability **Ingest only**, scope to **that one dataset only** (not "All datasets").
3. Add the token to each prod Key Vault as `shared-axiom-platform-ingest-token`.
4. Add the dataset name to each prod Key Vault as `shared-axiom-platform-dataset`.

## Post-merge: what happens automatically

| Deploy path | What rolls | When |
| --- | --- | --- |
| `.github/workflows/deploy-server.yml` (auto on push to main when `cmd/server/**` / `internal/**` change) | New server binary + Vector install via `az vm run-command` on every CP VM. populator runs at restart, fetches the KV token + dataset, vector starts shipping. | ~5 min after merge |
| `.github/workflows/build-worker-ami.yml` (auto on the same path triggers) | New worker AMI with the new binary + Vector pre-installed. First boot of any new worker fires the populator, then Vector. | ~20 min after merge for the build; existing workers stay on the old AMI until autoscaler-replaced. For an immediate fleet roll, manually run `deploy-worker.yml` (the workflow_dispatch hotfix path) — note: that workflow doesn't push the Vector configs the way deploy-server does, so it covers the binary but not Vector. The natural AMI roll covers Vector. |

**That's it.** No manual SSH steps required if both KV secrets are in place before merge.

If the KV secrets are missing at merge time: the populator exits 0, Vector starts but its Axiom healthcheck fails (no token, no dataset), events buffer to disk (256 MiB). Add the secrets to KV, run `systemctl restart populate-vector-env vector` on the affected hosts (or wait for natural restarts) — Vector picks up the new values and ships the backlog.

## Security notes

- The two prod Axiom tokens (`shared-axiom-ingest-token` for customer logs, new `shared-axiom-platform-ingest-token` for platform logs) should both be **ingest-only**. Verified via probe: an ingest-only token returns `403 token does not have access to resource: query with action: read` on the APL query endpoint, and `list datasets` returns only the scoped dataset. Audit this in the Axiom UI for both tokens before flipping the secret in KV.
- The platform token never crosses the host/guest boundary: it's only on worker/CP hosts in `/etc/opensandbox/vector.env` (mode 0600 root-only), never in `ConfigureLogship` RPCs, never in any rootfs image.
- The customer ingest token is reachable to a root-in-guest customer via agent process memory; this is a property of in-VM shipping not introduced by this PR. Mitigations (per-sandbox tokens, host-side sandbox shipping, server-side event stamping) are out of scope here.

## Test plan

- [x] `go test ./internal/obslog` — 7 tests including end-to-end Echo middleware round-trip (forwarded `X-Request-Id` lands on both handler logs and the access-log line) and a link-local filter test for `detectHostIP`.
- [x] `go build ./...` (modulo pre-existing Linux-only firecracker / agent build errors on `main`).
- [x] YAML lint on all three Vector configs and both modified workflows.
- [x] Validated end-to-end on a single-host dev: control plane + worker journald output is JSON with the envelope; Vector reads journald, ships to Axiom; an exec request through the proxy produces matching `request_id` events tagged `service=control-plane` AND `service=worker`.
- [x] Validated end-to-end on a **two-host** dev cluster (control-plane VM + worker VM in separate Azure VMs in the same VNet): the proxy hop crosses the network, the same `request_id` lands on both VMs' journald with different `host_ip` values, and both VMs' Vector instances ship cleanly to Axiom — no drops, no shipping errors.
- [x] Dev ingest-token capability probe confirmed ingest-only (no query, scoped to one dataset).

## Out of scope (follow-ups)

- DB logs: RDS doesn't expose host-side logs — needs RDS → CloudWatch → Axiom path, separate work.
- Migrate noisy `log.Printf` call sites in handler hot paths to `slog.InfoContext(ctx, ...)` so error lines emitted during a request carry the `request_id`.
- Extend `deploy-worker.yml` (the hotfix path) to also install/refresh Vector if you ever need to push a Vector change without going through a full AMI rebuild.

### DIFF
--- a/.github/workflows/build-worker-ami.yml
+++ b/.github/workflows/build-worker-ami.yml
@@ -94,6 +94,14 @@ jobs:
             deploy/ec2/build-rootfs-docker.sh \
             scripts/claude-agent-wrapper/
 
+      - name: Package Vector configs
+        # Bundled separately so Packer's file provisioner has a known
+        # tarball path (it can't reliably upload an arbitrary directory).
+        # install.sh on the builder extracts and installs Vector + the
+        # KV-token populator into the AMI.
+        run: |
+          tar czf /tmp/packer-vector-ctx.tar.gz -C deploy vector
+
       - name: Azure Login
         uses: azure/login@v2
         with:
--- a/.github/workflows/deploy-server.yml
+++ b/.github/workflows/deploy-server.yml
@@ -50,6 +50,13 @@ jobs:
       - name: Package web assets
         run: tar czf bin/web-dist.tar.gz -C web dist
 
+      - name: Package Vector configs
+        # Bundle deploy/vector/ for distribution to control-plane hosts.
+        # install.sh is idempotent: skips Vector install if already present,
+        # always refreshes config + restarts. Workers get the same files via
+        # the Packer image bake (build-worker-ami.yml).
+        run: tar czf bin/vector-deploy.tar.gz -C deploy vector
+
       - name: Azure Login
         uses: azure/login@v2
         with:
@@ -73,6 +80,14 @@ jobs:
             --container-name checkpoints \
             --name deploy/web-dist.tar.gz \
             --file bin/web-dist.tar.gz \
+            --overwrite true -o none
+
+          az storage blob upload \
+            --account-name ${{ secrets.AZURE_STORAGE_ACCOUNT }} \
+            --account-key "${{ secrets.AZURE_STORAGE_KEY }}" \
+            --container-name checkpoints \
+            --name deploy/vector-deploy.tar.gz \
+            --file bin/vector-deploy.tar.gz \
             --overwrite true -o none
 
           echo "Artifacts uploaded to blob storage"
@@ -120,6 +135,12 @@ jobs:
                   --container-name checkpoints --name deploy/web-dist.tar.gz \
                   --file /tmp/web-dist.tar.gz --overwrite true -o none
 
+                az storage blob download \
+                  --account-name ${{ secrets.AZURE_STORAGE_ACCOUNT }} \
+                  --account-key '${{ secrets.AZURE_STORAGE_KEY }}' \
+                  --container-name checkpoints --name deploy/vector-deploy.tar.gz \
+                  --file /tmp/vector-deploy.tar.gz --overwrite true -o none
+
                 # Stop, replace binary, extract frontend
                 systemctl stop opensandbox-server
                 sleep 1
@@ -136,6 +157,17 @@ jobs:
                 fi
 
                 systemctl start opensandbox-server
+
+                # Install/refresh Vector + the KV token populator. install.sh
+                # is idempotent (skips Vector install if already present,
+                # always refreshes config + reloads). populate-vector-env
+                # fetches AXIOM_PLATFORM_TOKEN from KV via the VM's managed
+                # identity at every boot (and on this restart).
+                mkdir -p /tmp/vector-deploy
+                tar xzf /tmp/vector-deploy.tar.gz -C /tmp/vector-deploy
+                bash /tmp/vector-deploy/vector/install.sh control-plane
+                rm -rf /tmp/vector-deploy /tmp/vector-deploy.tar.gz
+
                 echo 'Deploy complete'
               " -o none
 
--- a/.gitignore
+++ b/.gitignore
@@ -65,4 +65,5 @@ delme
 deploy/azure/.dev-env-state-*
 deploy/azure/.dev-env-secrets*
 !deploy/azure/.dev-env-secrets.example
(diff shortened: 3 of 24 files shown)


breardon2011 (approved):
approve

PR #242, opened 2026-05-13
hotfix(logship): sandbox-logs endpoint was returning cross-tenant data

motatoes:
## Summary

`GET /api/sandboxes/:id/logs` and `GET /api/dashboard/sessions/:id/logs` were returning logs from **every** sandbox in `oc-sandbox-logs` for the requested time window, not just the requested one. A caller hitting the endpoint for their own sandbox got back other tenants' sandbox logs interleaved with their own in the SSE stream.

## Root cause

Axiom's `/v1/datasets/<dataset>/query` endpoint silently ignores body fields it doesn't recognize. The handler was POSTing:

\`\`\`json
{
  "apl": "['oc-sandbox-logs'] | where sandbox_id == \"sb-X\" | ...",
  "startTime": "...",
  "endTime": "..."
}
\`\`\`

The endpoint has no \`apl\` field — that key was dropped, no filter applied, and the response contained every event in the dataset for the window. The response shape (\`{matches: [...]}\`) happens to match what the parser expected, so the handler returned events successfully but with no filtering.

Introduced in \`0da0bccd "fix(logship): SDK route + Axiom query format"\` — that commit's intent was to fix a 404 from a different APL endpoint by switching to the per-dataset endpoint. It "worked" (returned 200s) but "worked" meant "applied no filter".

## Fix

Switch to Axiom's native \`filter\` object on the same per-dataset endpoint. Same URL, same response parser, just a corrected request body:

\`\`\`json
{
  "startTime": "...",
  "endTime":   "...",
  "filter": {
    "op": "and",
    "filters": [
      {"op": "==", "field": "sandbox_id", "value": "<id>"},
      {"op": "contains", "field": "line", "value": "<q>"},
      {"op": "or", "filters": [{"op": "==", "field": "source", "value": "..."}, ...]}
    ]
  },
  "limit": 1000,
  "order": [{"field": "_time"}]
}
\`\`\`

Axiom applies the filter server-side, so the \`sandbox_id\` predicate runs before any events leave the dataset.

\`toAPL\` removed; \`toFilter\` + \`toRequest\` replace it. The \`q\` parameter no longer needs APL-string-literal escaping (Axiom's \`contains\` operator takes the value as a JSON string), so \`q.text = v\` instead of \`strings.ReplaceAll(v, '"', '\\"')\`. Control-char rejection in \`q\` is kept as defense in depth.

## Tests

8 tests, all passing. Two are load-bearing:

- \`TestFilter_AlwaysFiltersSandboxID\` — for every combination of caller-supplied query params (including injection attempts in \`q\`), the rendered filter has \`sandbox_id == <url-path-value>\` as the **first conjunct under the top-level AND**. Break this and the auth model breaks.
- \`TestRequest_BodyShape\` — asserts the marshaled JSON body must **not** contain an \`apl\` field. This is the regression test for the exact bug this PR fixes; if someone re-adds an \`apl\` field in the future, this fails before merge.

## Pre-merge

- [ ] Verify against a prod query token: hit the patched endpoint against a known multi-sandbox dataset and confirm the response only contains events for the requested \`sandbox_id\` (I can't do this from local — no prod query token here).
- [ ] Consider rotating \`shared-axiom-query-token\` after merge, since it has been queryable cross-tenant for a while (anyone who held it could have read any tenant's logs by changing the \`sandbox_id\` URL — though that required a valid org auth too).

## Out of scope

- Soft-delete vs hard-delete of \`sandbox_sessions\` rows on destroy (open question: can users fetch logs of a terminated sandbox via this endpoint? depends on whether the session row persists).
- CLI \`oc logs <sandbox-id>\` command — separate PR.
- LogsPanel in the web UI is **not** removed (still active); confirm what UX issue triggered the original "we took it out" memory before changing anything UI-side.

### DIFF
--- a/internal/api/sandbox_logs.go
+++ b/internal/api/sandbox_logs.go
@@ -19,12 +19,37 @@ import (
 )
 
 // allowedSources is the closed set of source values a client may filter
-// on. Anything else gets a 400 — never fed into the APL string.
+// on. Anything else gets a 400 — never fed into the query.
 var allowedSources = map[string]struct{}{
 	"var_log":     {},
 	"exec_stdout": {},
 	"exec_stderr": {},
 	"agent":       {},
+}
+
+// WARNING: Axiom's /v1/datasets/<dataset>/query endpoint silently
+// ignores unknown body fields. Sending `{"apl": "..."}` returns 200
+// with every event in the dataset (no filter applied). Always send the
+// `filter` field, and verify against a real response when changing the
+// body shape.
+type queryFilter struct {
+	Op      string        `json:"op"`
+	Field   string        `json:"field,omitempty"`
+	Value   any           `json:"value,omitempty"`
+	Filters []queryFilter `json:"filters,omitempty"`
+}
+
+type queryOrderField struct {
+	Field string `json:"field"`
+	Desc  bool   `json:"desc,omitempty"`
+}
+
+type queryRequest struct {
+	StartTime string            `json:"startTime"`
+	EndTime   string            `json:"endTime"`
+	Filter    queryFilter       `json:"filter"`
+	Limit     int               `json:"limit,omitempty"`
+	Order     []queryOrderField `json:"order,omitempty"`
 }
 
 // getSandboxLogs streams sandbox session logs as Server-Sent Events.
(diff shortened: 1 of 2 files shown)


breardon2011 (approved):
approve

PR #243, opened 2026-05-13
feat(logship): oc logs CLI, re-enable UI button, support stopped sandboxes

motatoes:
## Summary

Three connected pieces that round out the sandbox-logs feature:

### 1. `oc logs <sandbox-id>` CLI command

New file `cmd/oc/internal/commands/logs.go`. Live tail by default; flags mirror the API:

```
oc logs sb-abc                              # live tail
oc logs sb-abc --no-tail                    # one historical batch, then exit
oc logs sb-abc --since 30m                  # Go duration or RFC3339
oc logs sb-abc --grep error                 # server-side substring filter
oc logs sb-abc --source exec_stdout,exec_stderr
oc logs sb-abc --json                       # raw event JSON, one per line
```

Streams SSE from `/api/sandboxes/:id/logs` via a new `client.GetStream` helper (mirrors the existing `PostStream`). Formats events one-per-line with a source-coloured column (`var_log` cyan, `exec_stdout` gray, `exec_stderr` red, `agent` yellow), or emits raw JSON in `--json`.

### 2. Re-enable the Logs button in the web UI

Reverses #232 ("hide logs for now") — re-imports `LogsPanel`, re-adds the `showLogs` state, and re-renders the button + panel block. **One deliberate change**: the button now lives **outside** the `session.status === 'running'` gate, so users can read logs of stopped / hibernated / errored sandboxes too.

### 3. State-aware lifecycle handling in the handler

New `applySessionState(q, status, stoppedAt)` in `internal/api/sandbox_logs.go`. For non-running sandboxes (`stopped`, `error`, `hibernated`, `migrating`):

- `tail` forced to `false` — no new events will arrive once the sandbox is gone, polling is wasted work.
- `until` capped at `stoppedAt + 30s` grace — gives the in-VM shipper time to flush its last batch, then stops the query window so we don't ask Axiom for a void.

Pure function (no DB dependency), unit-tested independently. The DB-side enabler is already in place: `sandbox_sessions` rows are soft-deleted (status flipped, `stopped_at` set, row preserved), and `GetSandboxSessionInOrg` already returns those rows.

## Tests

| Suite | What it covers |
| --- | --- |
| `TestApplySessionState_*` (4 tests) | running passthrough; stopped/error/hibernated/migrating force `tail=false`; `until` cap behaviour (unset, tighter, future, no-stoppedAt) |
| `TestStreamSandboxLogs_*` (4 tests) | SSE parsing in formatted + JSON modes; comment/keepalive skip; malformed-payload tolerance |
| Existing `TestAPL_*` (5 tests) | Unchanged — still passing |

```
ok  github.com/opensandbox/opensandbox/internal/api               2.297s
ok  github.com/opensandbox/opensandbox/cmd/oc/internal/commands   0.900s
```

## Relationship to PR #242

This PR is **independent** of #242 (the cross-tenant filter hotfix). Both touch `internal/api/sandbox_logs.go` but in non-overlapping regions:

- #242 changes `queryAxiom` body shape (apl → filter) and replaces `toAPL`.
- This PR adds `applySessionState` in front of `queryAxiom` and re-enables UI surface.

They will merge cleanly in either order. **Recommended order is #242 first** — until that lands, all queries leak cross-tenant, so the new `oc logs` CLI and re-enabled UI button would surface the leak more visibly.

## Out of scope

- No change to `GetSandboxSessionInOrg` / `sandbox_sessions` lifecycle — rows are already soft-deleted, the DB side already supports post-mortem reads. The narrowing this PR adds is the query-side complement (don't tail a dead sandbox).
- No e2e CLI test — would need a live Axiom dataset + control plane. Unit tests cover the SSE parser; the network path is the same shape as `oc agent chat`'s streaming.

### DIFF
--- a/cmd/oc/internal/client/client.go
+++ b/cmd/oc/internal/client/client.go
@@ -170,6 +170,17 @@ func (c *Client) Post(ctx context.Context, path string, body, result interface{}
 	return nil
 }
 
+// GetStream performs a GET request and returns the raw response so the
+// caller can stream chunks (e.g. SSE log feeds). Caller closes resp.Body.
+func (c *Client) GetStream(ctx context.Context, path string) (*http.Response, error) {
+	req, err := http.NewRequestWithContext(ctx, "GET", c.baseURL+path, nil)
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("Accept", "text/event-stream")
+	return c.do(req)
+}
+
 // PostStream performs a POST request and returns the raw response so the
 // caller can stream chunks (e.g. SSE / chat-completions). Caller is
 // responsible for closing resp.Body.
--- a/cmd/oc/internal/commands/logs.go
+++ b/cmd/oc/internal/commands/logs.go
@@ -1,0 +1,181 @@
+package commands
+
+import (
+	"bufio"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/url"
+	"os"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/opensandbox/opensandbox/cmd/oc/internal/client"
+	"github.com/spf13/cobra"
+)
+
+var logsCmd = &cobra.Command{
+	Use:   "logs <sandbox-id>",
+	Short: "Stream a sandbox's session logs (live tail by default)",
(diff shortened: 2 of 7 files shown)


breardon2011 (approved):
approved

PR #244, opened 2026-05-13
metrics: ship platform metrics to Axiom via Vector

motatoes:
## Summary

- Adds platform-side observability that mirrors the existing log path (#240) and routes through its own Axiom dataset.
- New Prom histograms for HTTP request latency, sandbox create / checkpoint / hibernate / wake durations, plus failure-reason counters and worker resource gauges (disk, memory total/avail/allocated, CPU pressure via PSI with loadavg fallback).
- Vector gains a `prometheus_scrape` source + `metric_to_log` envelope transform + a separate `axiom_metrics` sink targeting a new `${AXIOM_PLATFORM_METRICS_DATASET}` so histogram-bucket volume doesn't share retention/cost with platform logs.
- Populator fetches two new KV secrets soft-optionally — if absent, logs keep flowing and the metrics sink buffers to disk until the operator provisions them.

## Why two datasets, not one

Histograms with N buckets emit ~N+2 events per series per scrape. `HTTPRequestDuration{method, path, status}` alone is plausibly bigger than current log volume on a busy worker. Mixing them into `oc-platform-logs` means a single retention policy, a single ingest budget, and one cardinality spike taking down incident-forensics queries. Two datasets gives independent retention/cost/rollout and is strictly additive (the metric_to_log transform stamps the same envelope so cross-dataset joins still work).

## Field shape in Axiom

Each metric event carries:
- `event_type = \"metric\"` (discriminator from log events)
- `metric_name` (Prom name; namespaced to dodge `name`-field clashes in log payloads)
- `tags` (Prom labels: region, worker_id, template, status, source, ...)
- `service`, `service_id`, `cell_id`, `region`, `host_ip`, `hostname` (same envelope as logs)
- gauge/counter/histogram value fields from Vector's `metric_to_log`

## Rollout

1. Land this PR → AMI rebake on merge (workflow already triggers on `deploy/vector/**`).
2. Workers/control-plane boot with new Vector configs. Logs ship as before. Metrics sink buffers to disk (healthcheck fails — visible but non-breaking).
3. Operator provisions `oc-platform-metrics` dataset + ingest token in Axiom; lands two KV secrets:
   - `shared-axiom-platform-metrics-ingest-token`
   - `shared-axiom-platform-metrics-dataset`
4. Next `populate-vector-env.service` run picks them up; metrics start flowing.

No coordinated worker restart needed — the soft-optional design lets each step land independently.

## Ports

- Worker `/metrics` stays on `:9091`.
- Control-plane `/metrics` newly served on `:9092` (dev-host parity — needs both binaries on one VM).

## Test plan

- [ ] `go build`/`go vet` clean on touched packages (verified locally for `GOOS=linux`).
- [ ] `vector validate deploy/vector/dev-host.yaml` (run on dev host — no `vector` binary locally).
- [ ] On dev-host: `curl -s localhost:9091/metrics | grep opensandbox_` and `curl -s localhost:9092/metrics | grep opensandbox_http_request_duration` show the new metrics.
- [ ] Drive sandbox lifecycle (create → checkpoint → hibernate → create-from-checkpoint) and confirm each histogram's `_count` increments.
- [ ] Force a checkpoint failure (fill data dir past 95%) and confirm `opensandbox_checkpoint_duration_seconds_count{status=\"failure\"}` and `opensandbox_checkpoint_failures_total{reason=\"disk_pressure\"}` both tick.
- [ ] Stress-loop host CPU and confirm `opensandbox_worker_cpu_pressure{window=\"avg10\",source=\"psi\"}` rises.
- [ ] In Axiom, query the new `oc-platform-metrics` dataset and confirm events appear with `region` / `service_id` populated. Confirm existing `oc-platform-logs` queries unchanged.

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -23,6 +23,7 @@ import (
 	"github.com/opensandbox/opensandbox/internal/controlplane"
 	"github.com/opensandbox/opensandbox/internal/crypto"
 	"github.com/opensandbox/opensandbox/internal/db"
+	"github.com/opensandbox/opensandbox/internal/metrics"
 	"github.com/opensandbox/opensandbox/internal/obslog"
 	"github.com/opensandbox/opensandbox/internal/observability"
 	"github.com/opensandbox/opensandbox/internal/proxy"
@@ -555,6 +556,13 @@ func main() {
 
 	addr := fmt.Sprintf(":%d", cfg.Port)
 	log.Printf("opensandbox: starting server on %s (mode=%s)", addr, cfg.Mode)
+
+	// Prometheus /metrics endpoint on a separate port. Different from the
+	// worker's :9091 so a dev-host (worker+server on one VM) doesn't collide.
+	// Vector scrapes this and ships to Axiom alongside platform logs.
+	metricsSrv := metrics.StartMetricsServer(":9092")
+	defer metricsSrv.Close()
+	log.Println("opensandbox: metrics server started on :9092")
 
 	go func() {
 		if err := server.Start(addr); err != nil {
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -166,6 +166,7 @@ func main() {
 			QEMUBin:         cfg.QEMUBin,
 			AgentBinaryPath: "/usr/local/bin/osb-agent",
 			AgentVersion:    AgentVersion,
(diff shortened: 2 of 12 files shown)


breardon2011 (approved):
Approve

PR #245, opened 2026-05-13
hotfix: cross-tenant leak still present — Axiom drops AND-wrapped filters

motatoes:
## Summary

**The cross-tenant leak that #242 was supposed to fix is still happening in prod.** I verified by re-running \`./scripts/test-sandbox-logs-isolation.sh\` against the patched server — it still reports the leak. Then I bypassed our handler entirely and probed Axiom directly with the **exact body shape** that the patched server sends, and got the smoking gun:

\`\`\`
test 1: wrapped AND, single conjunct, impossible sandbox_id → 50 events, mixed ids   ❌
test 2: flat ==,     no wrapper,        impossible sandbox_id →  0 events             ✅
test 3: wrapped AND, no filters at all                       → 50 events             ❌
test 4: wrapped AND, real sandbox_id                         → 50 events, mixed ids   ❌
\`\`\`

Axiom's per-dataset \`/v1/datasets/<dataset>/query\` endpoint silently drops AND-wrapped filter trees the same way it dropped the unknown \`apl\` field in the original bug. So #242's switch from \`{"apl":...}\` to \`{"filter":{"op":"and","filters":[...]}}\` was equivalent: another shape Axiom doesn't honor.

Only the **flat** predicate \`{"filter":{"op":"==","field":"sandbox_id","value":"<id>"}}\` actually narrows. Verified:

\`\`\`
flat ==, real sandbox_id sb-5b1696d4 → 13 events, all from sb-5b1696d4
\`\`\`

## Fix

- \`toFilter()\` now returns a flat \`==\` predicate. No \`.Filters\` slice population, no AND wrapper.
- New \`applyClientFilters()\` helper applies text-substring (q) and source-list (source) predicates **client-side**, after \`queryAxiom\` returns. Axiom doesn't compose those in the per-dataset filter shape; trying to AND them with sandbox_id would re-introduce the dropped-wrapper bug.
- Comment on \`queryFilter\` rewritten to warn explicitly about the wrapper trap so the next person isn't tempted to "improve" composition.

## Tests

| Test | Pins |
| --- | --- |
| **TestFilter_AlwaysFlatSandboxID** | The filter is always flat \`==\` on \`sandbox_id\`, no nested filters, regardless of caller input |
| **TestRequest_BodyShape** | The marshaled body must NOT contain \`"apl"\` (legacy #240 bug), \`"op":"and"\` (the #242 bug), or \`"op":"or"\` — any of those = Axiom drops the filter |
| **TestApplyClientFilters_*** (4 tests) | text substring, source list, composed AND of both, no-op pass-through |

All 11 tests pass.

\`\`\`
ok  github.com/opensandbox/opensandbox/internal/api  1.670s
\`\`\`

## Why client-side secondary filters?

The text and source predicates are pure UX, not security. Composing them server-side requires either:

1. AND-wrapping with the sandbox_id predicate → broken (this PR's reason).
2. Switching to the APL endpoint \`/v1/datasets/_apl\` → bigger surface change (URL + response parser).
3. Server-side post-filter → tiny diff, no Axiom-shape dependency.

Going with (3) for the hotfix; (2) is a clean follow-up if we want richer server-side query semantics.

The sandbox_id predicate stays Axiom-side. Tenant isolation does not depend on \`applyClientFilters\` (per its docstring) — even if a future bug skipped it, the events Axiom returned would already be sandbox-scoped.

## Verification before merge

Run \`./scripts/test-sandbox-logs-isolation.sh\` against any deployment that includes this commit. With the leak: it fails with cross-marker hits. With the fix: it prints \`PASS\`.

Direct Axiom probe:
\`\`\`bash
# Should return 0 events:
curl -X POST "https://api.axiom.co/v1/datasets/oc-sandbox-logs/query" \\
  -H "Authorization: Bearer $AXIOM_QUERY_TOKEN" \\
  -d '{"startTime":"...","endTime":"...","filter":{"op":"==","field":"sandbox_id","value":"sb-IMPOSSIBLE"}}'
\`\`\`

### DIFF
--- a/internal/api/sandbox_logs.go
+++ b/internal/api/sandbox_logs.go
@@ -28,10 +28,17 @@ var allowedSources = map[string]struct{}{
 }
 
 // WARNING: Axiom's /v1/datasets/<dataset>/query endpoint silently
-// ignores unknown body fields. Sending `{"apl": "..."}` returns 200
-// with every event in the dataset (no filter applied). Always send the
-// `filter` field, and verify against a real response when changing the
-// body shape.
+// ignores body it doesn't recognize — including AND-wrapped filter
+// trees. Empirically, `{"filter":{"op":"and","filters":[{...}]}}`
+// returns the unfiltered dataset (full leak); only a FLAT predicate
+// `{"filter":{"op":"==","field":"sandbox_id","value":"..."}}` actually
+// narrows. So toFilter() returns a single flat predicate on sandbox_id;
+// secondary filters (text search, source list) are applied client-side
+// in applyClientFilters after the query returns.
+//
+// The Filters slice is kept on the type for potential future
+// composition (e.g. if Axiom adds support, or if we move to /v1/datasets/_apl)
+// but is unset in the current shape.
 type queryFilter struct {
 	Op      string        `json:"op"`
 	Field   string        `json:"field,omitempty"`
@@ -137,6 +144,7 @@ func (s *Server) getSandboxLogs(c echo.Context) error {
 		writeSSEComment(c.Response(), "initial query failed")
 		return nil
 	}
+	rows = applyClientFilters(rows, q)
 	for _, ev := range rows {
 		writeSSEEvent(c.Response(), ev)
 	}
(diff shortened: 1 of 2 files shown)


breardon2011 (approved):
Approved

PR #247, opened 2026-05-13
fix(logship): stopped-sandbox event duplication + oc logs 404

motatoes:
## Summary

Two follow-up bugs from #243, found post-merge.

### 1. \`oc logs <sandbox-id>\` returns \`API error 404: Not Found\`

Path was constructed as \`/api/sandboxes/<id>/logs\`. But \`client.New()\` already appends \`/api\` to the base URL — so the actual request URL becomes \`/api/api/sandboxes/<id>/logs\` and Echo's router 404s.

Every other command in \`cmd/oc/internal/commands/\` uses the bare \`/sandboxes/...\` form (\`exec.go\`, \`sandbox.go\`); aligning logs to that.

### 2. Stopped sandboxes show each event 3+ times in the UI

\`\`\`
13:39:43.169 stdout 917900
13:39:43.282 stdout 4063628
13:39:44.999 stdout leakcheck-B-...
13:39:43.169 stdout 917900        ← REPEAT
13:39:43.282 stdout 4063628       ← REPEAT
13:39:44.999 stdout leakcheck-B   ← REPEAT
\`\`\`

Sequence: \`applySessionState\` flipped \`q.tail=false\` for non-running sandboxes → handler \`return nil\`-ed after the historical batch → SSE stream closed from the server → browser's \`EventSource\` interpreted the close as a disconnect and **auto-reconnected** → handler re-sent the same historical batch → repeat forever. Each reconnect appended duplicate events to LogsPanel state.

Fix:
- Don't flip \`q.tail\` in \`applySessionState\` — preserve the client's explicit intent (URL \`?tail=false\` / CLI \`--no-tail\`).
- In the handler, when the client requested \`tail=true\` but the sandbox is not running: skip the Axiom poll loop, but hold the SSE connection open with keepalive comments. EventSource doesn't reconnect on an open connection, so events render once and the stream stays quiet.
- CLI users who pass \`--no-tail\` still get close-on-completion (q.tail=false comes from the URL, independent of session state).

### Tests

Replaced \`TestApplySessionState_StoppedForcesHistorical\` with \`TestApplySessionState_PreservesTail\` — for every status × every clientTail value, q.tail comes through unchanged. The until-cap behaviour (already tested separately) is unchanged.

All 13 tests in \`internal/api\` pass; all 4 in \`cmd/oc/internal/commands\` pass.

### DIFF
--- a/cmd/oc/internal/commands/logs.go
+++ b/cmd/oc/internal/commands/logs.go
@@ -79,7 +79,9 @@ Examples:
 			qs.Set("limit", strconv.Itoa(limit))
 		}
 
-		path := "/api/sandboxes/" + url.PathEscape(sandboxID) + "/logs"
+		// client.New() already appends /api to baseURL — use the bare
+		// path here, matching the convention in exec/sandbox commands.
+		path := "/sandboxes/" + url.PathEscape(sandboxID) + "/logs"
 		if encoded := qs.Encode(); encoded != "" {
 			path += "?" + encoded
 		}
--- a/internal/api/sandbox_logs.go
+++ b/internal/api/sandbox_logs.go
@@ -159,6 +159,26 @@ func (s *Server) getSandboxLogs(c echo.Context) error {
 		return nil
 	}
 
+	// If the sandbox is not running, no new events will arrive — skip
+	// the Axiom poll loop entirely. But keep the SSE connection open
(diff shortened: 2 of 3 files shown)


breardon2011 (approved):
Approved

PR #248, opened 2026-05-14
metrics: stamp correct cell_id via shared-cell-id KV secret

motatoes:
## Summary

Vector's metrics envelope (`deploy/vector/control-plane.yaml:73`, same in `worker.yaml`) writes `cell_id = "${OPENCOMPUTER_CELL_ID:-unknown}"` **unconditionally**, so prom-scraped platform metrics ship to Axiom tagged `cell_id: unknown` whenever the env isn't set in Vector's process env — which is the current state in prod, since `OPENCOMPUTER_CELL_ID` isn't in Key Vault.

Logs aren't affected because the Go binary derives a fallback in `internal/config/config.go:284-285` (`cfg.CellID = cfg.Region + "-default"`) and the obslog handler stamps it on every record (`internal/obslog/obslog.go:79-80`), so journald lines arrive at Vector with `cell_id` already set and the `if !exists(.cell_id)` guard on the logs remap skips the env substitution entirely.

Wires `shared-cell-id` through the same dual-consumer pattern as the existing `shared-axiom-platform-*` secrets:

- **`internal/config/keyvault.go`** — adds `"shared-cell-id": "OPENCOMPUTER_CELL_ID"` to `secretMapping`. Keeps the file's documented role as the single source of truth for required `shared-*` KV secrets; lets the Go binary load it into its own env at startup so future app-side emitters get it without further plumbing.
- **`deploy/vector/populate-vector-env.sh`** — fetches `shared-cell-id` from KV and writes `OPENCOMPUTER_CELL_ID=...` into `/etc/opensandbox/vector.env`. Soft-fails to the populator's inherited shell env, then to literal `unknown`, so a missing KV secret doesn't break Vector boot — events just stay tagged `unknown` until the operator adds it.

## Per-cell action item

Add a `shared-cell-id` secret to each cell's Key Vault:
- `opencomputer-prod-kv` → `eastus2-default`
- dev KV → `westeurope-dev`

## Test plan

- [ ] On a prod-equivalent host, `systemctl restart populate-vector-env vector` and confirm `/etc/opensandbox/vector.env` now has `OPENCOMPUTER_CELL_ID=eastus2-default`
- [ ] In Axiom: `['oc-platform-metrics-prod'] | summarize count() by cell_id` shows `eastus2-default` for new events (still `unknown` for backlog)
- [ ] Per-cell dashboards under `terraform/axiom` (opencomputer-infra repo) populate without code change
- [ ] Confirm logs path unchanged — Go-side fallback still produces `eastus2-default` for hosts without the KV secret

### DIFF
--- a/deploy/vector/populate-vector-env.sh
+++ b/deploy/vector/populate-vector-env.sh
@@ -21,6 +21,7 @@
 #   shared-axiom-platform-dataset              → AXIOM_PLATFORM_DATASET        (required — logs)
 #   shared-axiom-platform-metrics-ingest-token → AXIOM_PLATFORM_METRICS_TOKEN  (optional — metrics)
 #   shared-axiom-platform-metrics-dataset      → AXIOM_PLATFORM_METRICS_DATASET (optional — metrics)
+#   shared-cell-id                             → OPENCOMPUTER_CELL_ID          (optional — cell tag)
 #
 # Logs secrets are hard-required: missing → exit 0 without writing env file,
 # Vector fails healthcheck and buffers to disk until the secret appears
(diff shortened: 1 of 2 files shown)


breardon2011 (approved):
Approve

PR #249, opened 2026-05-14
telemetry: fix cloud-init race + wire 3 unset dashboard gauges

motatoes:
Two related fixes that together get the prod dashboard's worker panels populating:

## 1. \`fix(vector): wait for cloud-final.service so worker.env is written\`

On freshly-booted Azure workers, \`populate-vector-env.service\` was racing cloud-init:
1. cloud-init's final stage base64-decodes the worker.env payload baked by the control plane (\`internal/compute/azure.go:651-655\`) and writes \`/etc/opensandbox/worker.env\` — including \`OPENSANDBOX_AZURE_KEY_VAULT_NAME=opencomputer-prod-kv\`.
2. The populator only declared \`After=network-online.target\`, so systemd was free to start it before cloud-final. Its \`EnvironmentFile=-/etc/opensandbox/worker.env\` (leading \`-\`) silently skipped the missing file.
3. The script saw \`OPENSANDBOX_AZURE_KEY_VAULT_NAME\` unset, logged \`"not set — skipping"\`, and exited 0.
4. Exit 0 means \`Restart=on-failure\` never fires → populator wedged for the rest of boot.
5. \`vector.service\` then failed to start because \`/etc/opensandbox/vector.env\` was never written and its \`${AXIOM_PLATFORM_TOKEN}\` / \`${OPENCOMPUTER_CELL_ID}\` substitutions had no values.

**Fix:** add \`After=cloud-final.service\` + \`Wants=cloud-final.service\` so the populator only runs once cloud-init has finished writing worker.env.

## 2. \`metrics: wire three dashboard gauges that were registered but never set\`

\`opensandbox_workers_total\`, \`opensandbox_auth_attempts_total\`, and \`opensandbox_sandboxes_active\` were defined and registered in \`internal/metrics/metrics.go\` but had zero \`.Set()\` / \`.Inc()\` call sites — the corresponding dashboard panels showed either "No data" or the misleading \`field 'tags.X' not found\` APL error (which is what APL returns on empty result sets when the query group-bys a tag column).

| Metric | Producer | Wiring location |
|---|---|---|
| \`opensandbox_workers_total\` | control plane | \`internal/controlplane/redis_registry.go\` — end of \`reconcileAndPrune()\` (10s tick), grouped by (region, status) |
| \`opensandbox_auth_attempts_total\` | control plane | \`internal/auth/oauth_handlers.go\` — each return path in \`HandleCallback\`, type=workos, result=success/failure |
| \`opensandbox_sandboxes_active\` | worker | \`internal/worker/resource_metrics.go\` (existing 30s tick) — driven by a new \`SandboxCounter\` interface satisfied by \`*qemu.Manager.ActiveSandboxesByTemplate()\` |

### Design notes

- **Gauge drift.** \`WorkersTotal\` and \`SandboxesActive\` both call \`.Reset()\` before emitting. Without it, a region that drained to zero (or a template that just had its last sandbox stop) would keep reporting its last non-zero value forever. The Reset → re-emit window is sub-microsecond; Vector scrapes on a 15-30s cadence so the race is negligible.
- **Heartbeat for sandboxes_active.** When no sandboxes are running, emit \`{template="", value=0}\` so the dashboard's \`summarize by tags.worker_id\` group-by has at least one row to chew on (renders \`0\` instead of "field not found").
- **Auth label cardinality.** Failure sub-reasons (missing_code, invalid_state, upstream_auth_failed, provision_failed) stay in logs rather than as metric label values — keeps \`result\` to a tight 2-value set.

## Test plan

- [ ] \`go build ./...\` + \`go vet ./...\` clean on touched packages (verified — pre-existing firecracker errors on main are unrelated).
- [ ] On a freshly-spawned worker built from this PR: \`journalctl -u populate-vector-env\` shows the populator ran AFTER cloud-init wrote worker.env and successfully fetched tokens from KV. \`systemctl is-active vector.service\` → \`active\`.
- [ ] On a control plane built from this PR: \`curl :9092/metrics | grep opensandbox_workers_total\` shows \`{region, status}\` rows matching live worker count, including \`draining\` if any.
- [ ] Successful + failed login → \`opensandbox_auth_attempts_total{type="workos",result="success"}\` and \`...{result="failure"}\` both visible.
- [ ] Worker with no sandboxes: \`curl :9091/metrics | grep opensandbox_sandboxes_active\` shows \`{template="",value=0}\`. Start a sandbox → row appears for that template.
- [ ] Dashboard panels (Active sandboxes, Workers, Auth attempts) populate after rollout instead of showing "No data" / "field not found".

## Rollout

- Worker AMI rebuild via \`build-worker-ami.yml\`, then fleet rollover. Control plane redeploy via the existing scp+systemctl-restart pattern in \`deploy/azure/create-opencomputer-prod.sh\`.
- For currently-running workers stuck in the populator-race state, the manual unblock is \`systemctl restart populate-vector-env.service vector.service\` (worker.env is already populated by now, so the populator will succeed on retry) — this is independent of the AMI/binary rebuild.

### DIFF
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -504,10 +504,12 @@ func main() {
 	// memory (sum of MemoryMB across running VMs), CPU pressure (PSI 'some'
 	// avg10/avg60/avg300, or loadavg/nproc fallback).
 	var allocator worker.MemoryAllocator
+	var sbCounter worker.SandboxCounter
 	if qemuMgr != nil {
 		allocator = qemuMgr
+		sbCounter = qemuMgr
 	}
-	worker.StartResourceMetricsTick(ctx, allocator, cfg.Region, cfg.WorkerID, cfg.DataDir, 30*time.Second)
+	worker.StartResourceMetricsTick(ctx, allocator, sbCounter, cfg.Region, cfg.WorkerID, cfg.DataDir, 30*time.Second)
 
 	// gRPC server (nil builder — template building via podman not needed for QEMU)
 	grpcServer := worker.NewGRPCServer(mgr, ptyMgr, execMgr, sandboxDBMgr, checkpointStore, sbRouter, nil, store)
--- a/deploy/vector/populate-vector-env.service
+++ b/deploy/vector/populate-vector-env.service
@@ -7,11 +7,20 @@
 # worker.env on workers, server.env on the control plane. EnvironmentFile
 # lines with a leading `-` are silently skipped if the file is absent, so
 # this unit works on both roles.
+#
+# Ordering note: on Azure workers, cloud-init's final stage writes
+# /etc/opensandbox/worker.env from a base64 payload baked by the control
+# plane (internal/compute/azure.go). Without `After=cloud-final.service`
+# the populator races cloud-init, reads an empty environ, exits 0 on
+# the "OPENSANDBOX_AZURE_KEY_VAULT_NAME not set" branch, and never
+# retries (Restart=on-failure doesn't fire on exit 0). Vector then can't
+# start because vector.env was never written. Ordering after cloud-final
+# guarantees worker.env exists by the time we read it.
 
 [Unit]
 Description=Populate /etc/opensandbox/vector.env from Key Vault
-After=network-online.target
-Wants=network-online.target
+After=network-online.target cloud-final.service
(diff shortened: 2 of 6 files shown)


breardon2011 (approved):
approved

PR #250, opened 2026-05-14
metrics: wire three dashboard gauges that were registered but never set

motatoes:
## Summary

\`opensandbox_workers_total\`, \`opensandbox_auth_attempts_total\`, and \`opensandbox_sandboxes_active\` were defined and registered in \`internal/metrics/metrics.go\` but had zero \`.Set()\` / \`.Inc()\` call sites in the codebase — the corresponding dashboard panels showed either "No data" or the misleading \`field 'tags.X' not found\` APL error (which happens on empty result sets when the query group-bys a tag column).

This PR wires each from the natural producer:

| Metric | Producer | Wiring location |
|---|---|---|
| \`opensandbox_workers_total\` | control plane | \`internal/controlplane/redis_registry.go\` — at end of \`reconcileAndPrune()\` (10s tick), grouped by (region, status) |
| \`opensandbox_auth_attempts_total\` | control plane | \`internal/auth/oauth_handlers.go\` — each return path in \`HandleCallback\`, type=workos, result=success/failure |
| \`opensandbox_sandboxes_active\` | worker | \`internal/worker/resource_metrics.go\` (existing 30s tick) — driven by a new \`SandboxCounter\` interface satisfied by \`*qemu.Manager.ActiveSandboxesByTemplate()\` |

## Design notes

- **Gauge drift.** \`WorkersTotal\` and \`SandboxesActive\` both call \`.Reset()\` before emitting. Without it, a region that drained to zero (or a template that just had its last sandbox stop) would keep reporting its last non-zero value forever. The Reset → re-emit window is sub-microsecond; Vector scrapes on a 15-30s cadence so the race is negligible.
- **Heartbeat for sandboxes_active.** When no sandboxes are running we still emit \`{template="", value=0}\` so the dashboard's \`summarize by tags.worker_id\` group-by has at least one row to chew on and renders \`0\` instead of "field not found".
- **Auth label cardinality.** Failure sub-reasons (missing_code, invalid_state, upstream_auth_failed, provision_failed) stay in logs rather than as metric label values — keeps \`result\` to a tight 2-value set.

## Test plan

- [ ] \`go build ./... && go vet ./...\` — both clean locally (verified).
- [ ] On a control plane built from this PR: \`curl :9092/metrics | grep opensandbox_workers_total\` shows \`{region, status}\` rows matching live worker count, including \`draining\` if any.
- [ ] Trigger a successful and a failed login → \`opensandbox_auth_attempts_total{type="workos",result="success"}\` and \`...{result="failure"}\` both visible.
- [ ] On a worker built from this PR with no sandboxes: \`curl :9091/metrics | grep opensandbox_sandboxes_active\` shows \`{template="",value=0}\`. Start a sandbox → row appears for that template.

## Rollout

Control-plane metrics ship with the next control-plane redeploy (\`deploy/azure/create-opencomputer-prod.sh:400-402\` scp+restart pattern). The sandboxes_active wiring lives in worker code so it needs a worker AMI rebuild (\`build-worker-ami.yml\`) and a fleet rollover, or the same SSH-overwrite shortcut for the existing fleet.

### DIFF
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -504,10 +504,12 @@ func main() {
 	// memory (sum of MemoryMB across running VMs), CPU pressure (PSI 'some'
 	// avg10/avg60/avg300, or loadavg/nproc fallback).
 	var allocator worker.MemoryAllocator
+	var sbCounter worker.SandboxCounter
 	if qemuMgr != nil {
 		allocator = qemuMgr
+		sbCounter = qemuMgr
 	}
-	worker.StartResourceMetricsTick(ctx, allocator, cfg.Region, cfg.WorkerID, cfg.DataDir, 30*time.Second)
+	worker.StartResourceMetricsTick(ctx, allocator, sbCounter, cfg.Region, cfg.WorkerID, cfg.DataDir, 30*time.Second)
 
 	// gRPC server (nil builder — template building via podman not needed for QEMU)
 	grpcServer := worker.NewGRPCServer(mgr, ptyMgr, execMgr, sandboxDBMgr, checkpointStore, sbRouter, nil, store)
--- a/internal/auth/oauth_handlers.go
+++ b/internal/auth/oauth_handlers.go
@@ -10,7 +10,16 @@ import (
 	"github.com/google/uuid"
 	"github.com/labstack/echo/v4"
 	"github.com/workos/workos-go/v4/pkg/usermanagement"
+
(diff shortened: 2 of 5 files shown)


motatoes:
Consolidated into #249 — both commits now land together so the worker fleet picks up the boot-order fix and the missing-gauge wirings in a single rollover.

PR #253, opened 2026-05-15
Fix wake billing memory tier

bibhuti-ssh:
## Summary
- Use the restored sandbox memory tier when resuming the wake billing scale event.
- Keep the existing 1024MB fallback for missing or legacy memory values.
- Share the fallback behavior with initial scale-event recording so create and wake use the same default.

## Why this matters
Hibernation ends the current sandbox_scale_events row. Wake starts a fresh row, but it was hard-coded to 1024MB. Any sandbox restored above 1GB would be under-recorded in sandbox_scale_events after wake until another scale event occurred. Usage reporting and billing aggregation read memory_mb from those scale events, so this fixes a real accounting/usage-reporting bug rather than just removing a TODO comment.

Disk is still passed as 0 intentionally so RecordScaleEvent inherits the previous disk size, matching the existing wake comment.

## Tests
- GOCACHE=/private/tmp/opencomputer-gocache GOPATH=/private/tmp/opencomputer-gopath go test ./internal/worker -count=1

### DIFF
--- a/internal/worker/grpc_server.go
+++ b/internal/worker/grpc_server.go
@@ -70,8 +70,8 @@ type LogshipConfigurator interface {
 type GRPCServer struct {
 	pb.UnimplementedSandboxWorkerServer
 	manager            sandbox.Manager
-	migrator           LiveMigrator      // optional, set if manager supports live migration
-	goldenRebuilder    GoldenRebuilder   // optional, set if manager supports golden rebuild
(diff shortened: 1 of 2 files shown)


motatoes:
hi @bibhuti-ssh thank you for this PR! we currently do not accept external contributions, if you have any issues or change requests please create an issue or reach out on slack and we will address it as soon as possible!

PR #254, opened 2026-05-15
vector: drop Wants=cloud-final from populator to break systemd ordering cycle

motatoes:
## Summary

PR #249 added both \`After=cloud-final.service\` AND \`Wants=cloud-final.service\` to \`populate-vector-env.service\`. The \`Wants=\` half pulled cloud-final into Vector's dependency graph and created an ordering cycle that systemd silently resolves by deleting \`vector.service/start\` — Vector never boots, no log, no error.

## Reproduction on prod

After #249 merged and a fresh worker rolled, observed on one prod worker:
- \`uptime\`: 1h 50m — fresh boot on the new AMI
- \`/etc/opensandbox/vector.env\`: fully populated (populator ran correctly)
- \`opensandbox-worker.service\`: active
- \`vector.service\`: **inactive (dead)** — never tried to start, zero journal entries this boot

In the early boot journal:
\`\`\`
cloud-final.service: Found dependency on vector.service/start
cloud-final.service: Job vector.service/start deleted to break ordering cycle
                     starting with cloud-final.service/start
\`\`\`

## Why \`After=\` is enough

\`After=cloud-final.service\` already gives the ordering needed to fix the original cloud-init race #249 was solving. \`Wants=\` adds a "pull into dep graph" semantic we don't actually need — cloud-final is a stock cloud-init target that's always present, no need to "want" it.

## Test plan

- [ ] Stage on a dev worker — replace \`/etc/systemd/system/populate-vector-env.service\` with this version, \`systemctl daemon-reload\`, reboot
- [ ] After reboot: \`systemctl is-active vector.service\` returns \`active\`
- [ ] Boot journal contains zero "ordering cycle" entries
- [ ] Metric events from worker show up in Axiom within 1–2 minutes

### DIFF
--- a/deploy/vector/populate-vector-env.service
+++ b/deploy/vector/populate-vector-env.service
@@ -10,26 +10,26 @@
 #
 # Ordering note: on Azure workers, cloud-init's final stage writes
 # /etc/opensandbox/worker.env from a base64 payload baked by the control
-# plane (internal/compute/azure.go). Without `After=cloud-final.service`
-# the populator races cloud-init, reads an empty environ, exits 0 on
-# the "OPENSANDBOX_AZURE_KEY_VAULT_NAME not set" branch, and never
-# retries (Restart=on-failure doesn't fire on exit 0). Vector then can't
-# start because vector.env was never written.
+# plane (internal/compute/azure.go). The populator races cloud-init and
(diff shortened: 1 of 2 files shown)


breardon2011 (approved):
approve

PR #260, opened 2026-05-18
vector: detach populator from boot when role env is missing

motatoes:
## Summary

- Make `populate-vector-env.sh` exit fast in **every** boot path. When `/etc/opensandbox/{worker,server}.env` is missing (Azure first boot, cloud-init not yet finished), it writes a stub `vector.env` so `vector validate` passes, then kicks a new companion unit `populate-vector-env-wait.service` to poll asynchronously. Boot proceeds in ~1s.
- The wait unit (not `WantedBy=multi-user.target`, so it cannot block boot) polls for the role env file for up to 30 min, then re-runs the populator (which now goes through the synchronous Key Vault path) and `systemctl reset-failed + restart vector.service` so the disk-buffered events flush with the real Axiom token.

## Why

`#257` shipped a 600s synchronous in-script wait. On Azure that deadlocked the boot graph: `cloud-final.service` is ordered `After=multi-user.target` on Ubuntu Azure images, and `cloud-final` is what writes `worker.env` — but `multi-user.target` couldn't reach active while the populator was waiting (`vector.service` Wants= populator, `multi-user.target` Wants= vector). Every new Azure worker was reaped at exactly 600s by `scaler.go`'s `pendingWorkerTTL=10min` and replaced, ad infinitum.

Reverted in #259. This PR is the proper fix.

## What prior attempts hit (full history in `populate-vector-env.sh` header)

| PR | Approach | Failure mode |
|----|----------|--------------|
| #249 | `After=cloud-final.service` | systemd cycle, `vector.service/start` silently deleted |
| #254 | `exit 1` + `Restart=on-failure` | vector's restart-burst burned `StartLimitBurst=5` in <2s |
| #256 | internal 90s in-script poll | multi-user blocked 90s, vector burned its own restart budget |
| #257 | internal 600s in-script poll | full boot deadlock, every Azure worker reaped at 10min |

## Test plan

- [ ] `bash -n` syntax check on both shell scripts (done locally)
- [ ] Build worker AMI on `opencomputer-deployment-test` from this branch
- [ ] Provision a worker from the new AMI in test Azure environment
- [ ] Confirm via `az vm run-command`:
  - [ ] `cloud-init status` reaches `done` (not stuck at `running`)
  - [ ] `populate-vector-env.service` exits 0 within a few seconds
  - [ ] `populate-vector-env-wait.service` is `activating`/`running`
  - [ ] `vector.service` reaches active with the stub env (axiom sink unhealthy + buffering — expected)
  - [ ] After `cloud-final` writes `worker.env`, waiter re-runs populator and vector restarts with real creds (axiom sink healthy)
  - [ ] Worker registers with control plane within `pendingWorkerTTL`
  - [ ] Worker logs appear in Axiom
- [ ] Confirm dev-host install path still goes through the synchronous branch (no waiter spawn)

### DIFF
--- a/deploy/vector/install.sh
+++ b/deploy/vector/install.sh
@@ -55,11 +55,16 @@ install -m 0644 "$CONFIG_SRC" /etc/vector/vector.yaml
 chown -R vector:vector /var/lib/vector
 
 # --- Install the KV → env-file populator (oneshot, runs Before=vector.service) ---
-# Both the script and the unit file are tracked in this dir so they roll
-# atomically with config changes.
+# Plus a companion wait unit that polls for worker.env asynchronously when
+# the main populator races ahead of cloud-init on Azure first boot. All
+# four files are tracked in this dir so they roll atomically with config
+# changes.
 install -m 0755 "$SCRIPT_DIR/populate-vector-env.sh" /usr/local/bin/populate-vector-env.sh
+install -m 0755 "$SCRIPT_DIR/populate-vector-env-wait.sh" /usr/local/bin/populate-vector-env-wait.sh
 install -m 0644 "$SCRIPT_DIR/populate-vector-env.service" \
     /etc/systemd/system/populate-vector-env.service
+install -m 0644 "$SCRIPT_DIR/populate-vector-env-wait.service" \
+    /etc/systemd/system/populate-vector-env-wait.service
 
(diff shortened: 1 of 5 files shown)


breardon2011 (approved):
Approve

PR #271, opened 2026-05-23
honor memoryMB on fork-from-checkpoint

breardon2011:
## Summary

Forks from checkpoints/templates silently ignored the caller's `memoryMB`
and always landed at the snapshot's recorded size. Customer reported this
as "always starts with 896 MiB usable" — that's the kernel-visible
memory for a snapshot taken at `-m 1024`. They worked around it by
manually scaling post-launch.

Root cause was in two places:

1. **API layer:** `POST /api/sandboxes/from-checkpoint/:id` bound only
   `envs`/`secretStore`/`metadata` from the request body, so any
   `memoryMB` the caller sent was dropped before reaching the worker.
2. **Worker layer:** Even when given a `cfg.MemoryMB`,
   `ForkFromCheckpoint` ignored it and launched at the snapshot's recorded
   base. QEMU live migration requires source/dest `-m` to match exactly,
   so a direct launch at the requested size isn't possible — but the
   running guest can be resized via virtio-mem.

## What changed

- **API:** `memoryMB` is now an accepted field on the fork body and
  propagates through `createFromCheckpointCore` → gRPC →
  `ForkFromCheckpoint`. Effective value comes back as `memoryMB` in the
  response so callers can detect when the request was clamped.
- **Worker:** After loading the migration state but **before** `Cont`,
  virtio-mem is plugged to the requested size (mirrors the existing
  `RestoreFromCheckpoint` and wake-from-hibernate patterns at
  `manager.go:3237` and `snapshot.go:507`). VM is paused at this point, so
  the kernel sees the full memory map on resume — no race between guest
  workloads and ongoing hotplug.
- **Clamps:**
  - **Floor** = snapshot's running memory (`meta.MemoryMB`, includes any
    virtio-mem plugged at snapshot time). Below-floor requests are
    refused: shrinking would OOM restored processes whose working set
    was sized for the snapshot.
  - **Ceiling** = QEMU `maxmem` from `buildQEMUArgs` (= base + virtio-mem
    pool, effectively 16 GB for any base ≤ 16 GB). Above-ceiling requests
    are clamped down.
  - In both cases the request is honored to the clamped value and the
    server logs a warning naming the inputs and reason. The effective
    value is returned in the response.
- **Proto:** `CreateSandboxResponse.memory_mb` added.

## Test plan

Verified end-to-end against dev (`https://dev.opensandbox.ai`), matrix:

| # | Snapshot base / total | Fork request | Response `memoryMB` | Guest `MemTotal` | Worker log |
|---|---|---|---|---|---|
| 1 | 1024 / 1024 | omit | 1024 | 917 MB | (baseline, no warn) |
| 2 | 1024 / 1024 | 4096 | 4096 | 4063 MB | `+3072MB total=4096` |
| 3 | 1024 / 4096 | 1024 | 4096 | 4063 MB | `ignored ... downscale would OOM` + `+3072MB total=4096` |
| 4 | 1024 / 4096 | 32768 | 16384 | 16646 MB | `clamped to hotplug ceiling 16384MB` + `+15360MB total=16384` |

Fork latency: ~2-3s for small hotplugs (typical 1→4 GB, 4→8 GB jumps);
~26s for the +15 GB plug in scenario 4. Plugging large amounts pre-resume
is genuinely expensive — most customers will see the small-hotplug path.

- [x] All 4 scenarios show the expected `memoryMB` in the API response
- [x] `cat /proc/meminfo` inside each fork confirms the guest actually
      sees the effective memory (within kernel-overhead margin)
- [x] Worker journal contains the floor/clamp warning lines for #3 and #4
- [x] `go build ./internal/qemu/... ./internal/api/... ./internal/worker/... ./proto/...` clean
- [x] vCPU count is unchanged on the fork path — savevm topology lock
      prevents adding cores; CPU "scaling" continues to flow through
      cgroup throttling post-fork as today


breardon2011:
Closing, added this functionality in another  PR

Issue #290, opened 2026-05-23
`doWake` invoked with request context — caller cancel strands waiters


ravidsrk:
**Type:** Robustness
**Severity:** Medium
**Component:** `internal/sandbox`
**Suggested labels:** `bug`, `area/sandbox`, `priority/medium`

## Summary
`SandboxRouter.Route` calls `r.doWake(ctx, sandboxID, entry)` with the *caller's* request context at line 359. `doWake` internally creates its own `context.WithTimeout(context.Background(), 60*time.Second)` for the actual wake, so the value isn't used for I/O — but if the caller's context is already cancelled by the time `doWake` is entered (or cancels during the brief window before `wakeCtx` is created), there is no actual cancellation hazard for the I/O. The real defect is the *signal* it sends to future maintainers and the deeper subtle behavior: the caller goroutine that entered the `StateHibernated → StateWaking` transition holds responsibility for completing the wake on behalf of every other waiter. If that goroutine is the one whose context gets cancelled and any future code inside `doWake` ever uses the passed `ctx` (instead of `wakeCtx`), other waiters block until their own contexts expire.

## Affected code
- `internal/sandbox/router.go:351-368` — `ensureRunning` transitions to `StateWaking`, then calls `doWake(ctx, ...)` synchronously on the entering goroutine
- `internal/sandbox/router.go:435-490` — `doWake` shadows the passed `ctx` with its own `wakeCtx` derived from `context.Background()`

## Current behavior
```go
// internal/sandbox/router.go:351-368
case StateHibernated, StateHibernating:
    // Transition to waking, start wake, other requests will queue
    entry.state = StateWaking
    entry.wakeCh = make(chan struct{})
    entry.wakeErr = nil
    entry.mu.Unlock()

    // Perform wake (only one goroutine does this)
    r.doWake(ctx, sandboxID, entry)   // <-- caller's ctx passed in

    // Check wake result
    entry.mu.Lock()
    err := entry.wakeErr
    entry.mu.Unlock()
    if err != nil {
        return fmt.Errorf("auto-wake failed for sandbox %s: %w", sandboxID, err)
    }
    return nil
```

```go
// internal/sandbox/router.go:435-490
func (r *SandboxRouter) doWake(ctx context.Context, sandboxID string, entry *sandboxEntry) {
    // Use a background context for the wake itself so it completes
    // even if the original request times out
    wakeCtx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
    defer cancel()
[...]

PR #311, opened 2026-05-23
chore(sandbox): make doWake's context independence explicit at the call site

ravidsrk:
Partial fix for #290 — see note on issue.

## Summary

`doWake` already ignored its `ctx` parameter and derived its own deadline from `context.Background()` (`internal/sandbox/router.go:438`), so wakes survive request-context cancellation. The intent was correct but invisible at the call site: readers see `r.doWake(ctx, ...)` and assume the request context is honored.

## Change

Two no-op-at-runtime cleanups:

```go
// Route() — make the intent local:
r.doWake(context.Background(), sandboxID, entry)

// doWake — rename to _ and document:
func (r *SandboxRouter) doWake(_ context.Context, sandboxID string, entry *sandboxEntry) {
    wakeCtx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
    ...
}
```

## Note on the issue scope

The issue's "caller cancel strands waiters" failure mode doesn't actually reproduce because doWake never observed the caller's context — the unused-parameter design already protected against it. This PR cleans up the misleading signature so future readers don't get the wrong impression and accidentally start using the parameter.

If the maintainers want a stronger fix (e.g., removing the parameter entirely), I can follow up — it'd be a small contract change.

### DIFF
--- a/internal/sandbox/router.go
+++ b/internal/sandbox/router.go
@@ -355,8 +355,13 @@ func (r *SandboxRouter) ensureRunning(ctx context.Context, sandboxID string) err
 		entry.wakeErr = nil
 		entry.mu.Unlock()
 
-		// Perform wake (only one goroutine does this)
-		r.doWake(ctx, sandboxID, entry)
+		// Perform wake (only one goroutine does this). Pass a background
+		// context — doWake derives its own deadline from
+		// context.Background() internally so it survives the caller's
+		// request context being cancelled. Passing the request context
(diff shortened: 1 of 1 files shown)


motatoes:
Hello @ravidsrk thank you for these PRs, unfortunately we are not able to accept PRs from external contributors at this time, we should have made it more clear in our docs, appologies for the misunderstanding

PR #353, opened 2026-06-05
Add Burst Sandbox edge routing

motatoes:
## Summary
- routes Burst Sandbox creates through api-edge using burst: true
- adds Cloudflare Worker config for BURST_CELL_ID routing support

## Verification
- npx tsc --noEmit in cloudflare-workers/api-edge


breardon2011 (approved):
Approved

PR #395, opened 2026-06-19
ci: require SDK version bump + lockfile sync on SDK source PRs

ZIJ:
**Why:** `publish-ts-sdk.yml` only publishes the TypeScript SDK when `package.json`'s
version differs from what's on npm. So a PR that changes `sdks/typescript/src` *without*
bumping the version merges cleanly but never publishes — the changed code silently never
ships. That's exactly what happened with #393 (the `connectSession` fix), which needed a
separate follow-up bump PR (#394) to actually release.

**What:** a `pull_request` check on `sdks/typescript/**` that:
- exempts docs/README-only changes (only fires when `sdks/typescript/src` actually changed),
- fails if `package.json`'s version equals the base branch's version,
- fails if `package-lock.json`'s version is out of sync with `package.json`.

So an SDK change can't merge without being releasable.

### DIFF
--- a/.github/workflows/pr-sdk-version-bump.yml
+++ b/.github/workflows/pr-sdk-version-bump.yml
@@ -1,0 +1,50 @@
+name: SDK version bump
+
+# Guards against shipping TypeScript SDK source changes without a version bump.
+# publish-ts-sdk.yml only publishes when package.json's version differs from what's on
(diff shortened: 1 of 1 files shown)


ZIJ:
Decided this is too minor to maintain a dedicated workflow. Relying on a documented rule + review discipline instead: keep the SDK version bump and package-lock sync in the same PR as any sdks/typescript/src change.

PR #407, opened 2026-06-23
docs: contract-consistency fixes (messaging shape, ids, idempotency, params)

ZIJ:
Docs side of a docs/API/SDK contract review. The SDK-side fixes are in #406.

- **P1b — messaging**: rewrote the "envelope" section to **Message input** + a table of where each field actually lands. A `*.message` event `body` is `{ text, raw? }`; `source`/`conversation`/`refs` go to the event's top-level `refs` and the sender to `actor` — the docs previously described a single normalized envelope body the API doesn't emit.
- **P2b — POST /agents idempotency**: it's distinguished by **status code** (201 create / 200 idempotent hit / 409 conflict), not a `created:false` body field (the API/SDK don't return one).
- **P2/P3 — verbatim scope**: only session `metadata` and an event's `raw` pass through verbatim; event `body`/`refs` are camelCased like everything else (matches SDK `normalize`).
- **P2a — SSE filtering**: documented that the live stream filters by `level`+`after` server-side; `type` on a stream is applied client-side by the SDK's `events({ type })` (the matching SDK change is in #406).
- **P3a — example ids**: `ses_…` (not `sess_…`) everywhere; `client_token` shown as a JWT (`eyJ…`).
- **P3b — params**: documented `GET /sessions?cursor=` and `GET …/deliveries?after=&limit=`.

Companion: **#406** (SDK) does the matching `verifyWebhook` normalize, stream `type` client-side filter, and `credentials.create({ isDefault })`.

### DIFF
--- a/docs/agent-sessions/api-reference.mdx
+++ b/docs/agent-sessions/api-reference.mdx
@@ -11,11 +11,11 @@ Each section toggles between the **TypeScript** SDK (`@opencomputer/sdk`) and **
 import { OpenComputer, connectSession } from "@opencomputer/sdk";
 
 const oc = new OpenComputer({ apiKey: process.env.OPENCOMPUTER_API_KEY! });
-const session = await oc.sessions.get("sess_...");        // server, org key
+const session = await oc.sessions.get("ses_...");        // server, org key
 // or, in a browser:  await connectSession({ sessionId, clientToken })
 ```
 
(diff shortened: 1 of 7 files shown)


ZIJ:
Folded into #406 — the docs describe the SDK behavior in that PR (SSE client-side type filter, verifyWebhook normalize) and both touched webhooks.mdx, so they should land together. Closing; branch deleted.

PR #410, opened 2026-06-24
Sandbox webhooks

ZIJ:
## Sandbox lifecycle webhooks

Adds sandbox lifecycle webhooks using the existing edge event pipeline plus Svix managed delivery.
This replaces the earlier regional-control-plane dispatcher/ledger design: the control plane emits
lifecycle facts, Cloudflare owns webhook management and handoff, and Svix owns fan-out, retries,
signing, SSRF protection, delivery logs, and resend.

### What changed

- **api-edge owns the public webhook API:** `/api/webhooks` create/list/get/update/delete,
  `/test`, `/deliveries`, `/redeliver`, and the D1 destination index that maps OC destination ids to
  Svix applications/endpoints.
- **events-ingest owns delivery handoff:** accepted `sandbox.*` events are written to D1, gated by
  `orgs.has_webhooks`, then synchronously handed to Svix before returning 202. Transient Svix errors
  return 503 so the existing `EventForwarder` retries the Redis-stream batch with the same stable
  event id / idempotency key.
- **The control plane stays thin:** it keeps only the transactional lifecycle outbox for CP-origin
  events, drained to the existing cell event stream. It no longer owns endpoint management,
  delivery rows, retry timers, signing, SSRF checks, or a Svix client.
- **Inline-on-create is preserved:** `POST /api/sandboxes { webhooks: [...] }` registers a
  sandbox-scoped endpoint through edge `/internal/webhooks/register` before the sandbox emits
  `created`.
- **Docs + TypeScript SDK are updated** for sandbox lifecycle webhooks, inline create fields,
  Svix/Standard Webhooks headers, typed lifecycle events, delivery helpers, and the shared
  `verifyWebhook` helper.

### Public contract

- Events: `sandbox.created`, `sandbox.ready`, `sandbox.hibernated`, `sandbox.resumed`,
  `sandbox.stopped`, `sandbox.migrated`, `sandbox.checkpoint.created`, `sandbox.forked`,
  `sandbox.scaled`, `sandbox.preview_url.changed`.
- Delivery is **at least once after edge/Svix handoff**, not strictly ordered. Consumers should dedupe
  on the stable Svix delivery id and treat each event independently.
- Signing uses Standard Webhooks via Svix headers: `svix-id`, `svix-timestamp`, `svix-signature`.
- Per-destination `metadata` is delivered as Svix endpoint custom headers. The sandbox metadata
  snapshot was intentionally removed from event bodies; callers can fetch sandbox metadata by id.
[...]

PR #412, opened 2026-06-24
ci: events-ingest deploy + D1 schema apply workflows

ZIJ:
Closes the two **manual gaps** the PR #410 merge surfaced (webhooks shipped, but two pieces had no automation):

## 1. `deploy-events-ingest.yml` — events-ingest had no CI deploy
api-edge and the control plane auto-deploy on merge; **events-ingest did not** (only a `workflow_dispatch` `deploy-worker.yml` for the sandbox AMI). So on #410's merge its prod worker kept running old code until it was deployed by hand. This mirrors `deploy-api-edge.yml`: auto-deploys to prod on merge touching `cloudflare-workers/events-ingest/**` or `shared/**`, plus a `workflow_dispatch` prod/dev target. (No dashboard-asset build and no test/tsc step — events-ingest has neither.)

## 2. `apply-d1-schema.yml` — D1 schema was applied by hand
CI doesn't run `d1 execute`, so the webhooks D1 schema (`has_webhooks`, `webhook_destinations`, `webhook_idempotency`) was applied manually with local wrangler. This adds a **`workflow_dispatch`** job to apply a chosen schema file to the prod/dev D1 from CI.

**Why manual-trigger (not on-push):** D1 has no native auto-migration, and the `schema_phase*.sql` files use non-idempotent `ALTER TABLE ADD COLUMN` (SQLite has no `IF NOT EXISTS` for columns), so re-applying on every push would fail. This is the safe, immediately-useful version (no local wrangler/account juggling).

### Follow-up (intentionally not in this PR)
Fully-automatic on-merge D1 application would require adopting **`wrangler d1 migrations`** (a `d1_migrations` tracking table + numbered migrations) **plus a one-time baseline** of the already-populated prod D1 (mark existing schema as applied so the ALTERs don't re-run). That's a delicate change on a live DB and deserves its own review — happy to do it next if we want it.

### DIFF
--- a/.github/workflows/apply-d1-schema.yml
+++ b/.github/workflows/apply-d1-schema.yml
@@ -1,0 +1,70 @@
+name: Apply D1 Schema
+
+# CI-assisted application of a D1 schema file to a target database, so schema
+# changes don't require a local wrangler + account juggling.
+#
+# This is workflow_dispatch (NOT push-triggered) on purpose: D1 has no native
+# auto-migration, and the phase schema files use non-idempotent `ALTER TABLE ADD
+# COLUMN` statements (SQLite has no `IF NOT EXISTS` for those), so blindly
+# re-applying them on every push would fail. Pick the DB + file and run it once.
+#
+# A fully-automatic on-merge system would require adopting `wrangler d1
+# migrations` (a tracking table + numbered migrations) plus a one-time baseline
+# of the already-populated prod DB. Tracked as a follow-up — see the PR.
+
+on:
+  workflow_dispatch:
+    inputs:
+      target:
+        description: 'D1 database'
+        required: true
+        default: 'prod'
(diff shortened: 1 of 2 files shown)


breardon2011 (approved):
approve

motatoes:
I've added the ingest deployment workflow here to merge it in https://github.com/diggerhq/opencomputer/pull/423/changes#diff-568ba180ef62c7170dc213e88aaa7f9826c29abc73c8fe1b152a779d35ca1e28

the migrations im not sure if this loop will work reliably, we need some kind of forward background framework to apply them - not sure whats there  in cloudflare world

ZIJ:
back-and-forth because I merged as soon as I saw Brian's comment, then saw Mo's

motatoes:
🤦‍♂️

PR #427, opened 2026-06-26
Auto-promote disk-only checkpoints by default

motatoes:
## Summary
- Default disk-only checkpoints to background full promotion when promoteToFull is omitted
- Preserve the hidden explicit promoteToFull=false override for advanced/manual testing
- Hide the CLI promotion flag from normal help and remove it from public Python SDK docs
- Add API and TypeScript SDK regression coverage

## Tests
- go test ./internal/api -run TestShouldPromoteCheckpointDefault
- npm test -- --run src/sandbox.test.ts
- python3 -m py_compile opencomputer/sandbox.py


breardon2011 (approved):
approve

PR #443, opened 2026-06-28
Fix dialog dropdown outside-dismiss handling

ZIJ:
## Summary
- marks portaled Select and DropdownMenu content as floating layers
- prevents DialogContent and SheetContent from treating nested floating-layer interactions as outside-dismiss events
- handles the Radix Select edge case where `disableOutsidePointerEvents` retargets a second same-position click to the overlay while the select layer is still open/dismissing
- keeps the guard alive through the pointerdown → pointerup → click sequence instead of clearing it on `setTimeout(0)`

## Testing
- `PATH="$HOME/.nvm/versions/node/v22.13.0/bin:$PATH" npm run typecheck`
- `PATH="$HOME/.nvm/versions/node/v22.13.0/bin:$PATH" npx eslint src/components/form.tsx src/components/ui/dialog.tsx src/components/ui/dropdown-menu.tsx src/components/ui/floating-layer.ts src/components/ui/sheet.tsx`
- `PATH="$HOME/.nvm/versions/node/v22.13.0/bin:$PATH" npm run build`
- Headless Chrome against `http://localhost:3000` with only `/api/dashboard/*` mocked in-page: verified Agents, Sessions, and Credentials dialogs stay open after the same-position second click on nested selects, including after a 1s wait.
- Earlier preview-mode check also verified selecting nested select items keeps the parent dialog open.

## Note
- Full `npm run lint` still fails on existing `src/pages/SessionDetail.tsx:82` (`react-hooks/set-state-in-effect`), unrelated to this change.

### DIFF
--- a/.agents/work/web-known-issues.md
+++ b/.agents/work/web-known-issues.md
@@ -5,56 +5,47 @@ something is reproducible but not yet fixed, with enough detail to resume cold.
 
 ---
 
-## Dialog closes when interacting with a Select/dropdown inside it
+## Dialog closes when interacting with a Select/dropdown inside it — RESOLVED
 
-**Status:** open · **Severity:** minor–moderate UX (loses in-progress form input) · **Filed:** 2026-06-28
+**Status:** resolved (PR #443) · **Filed:** 2026-06-28 · **Fixed:** 2026-06-29 · **Severity:** minor–moderate UX (lost in-progress form input)
 
-**Symptom.** Open a Radix `Select` (or any dropdown) rendered *inside* a `Dialog`, then
-click an item — notably opening the dropdown a second time and clicking again — and the
-**whole Dialog closes** instead of just the dropdown. Confirmed **generic to every modal
-with a dropdown**, not specific to one screen (first reported on the agent-create dialog's
-credential picker; reproduced elsewhere).
+**Symptom.** Open a Radix `Select` inside a `Dialog`, then click the trigger a *second*
+time to close it — and the **whole Dialog closes** instead of just the menu. Generic to
+every modal with a dropdown (first hit on the agent-create credential picker).
 
-**Environment.** `radix-ui ^1.6.0` (the unified package), React 19, vite dev on `:3000`,
-`web/`. Affected primitives: `web/src/components/ui/dialog.tsx` (Dialog) +
-`web/src/components/form.tsx` (`Select`, `position="popper"`). Surfaces: any dialog with a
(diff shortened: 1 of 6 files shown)


ZIJ:
Reworked the fix to keep the same Radix-blessed mechanism but make the *condition* precise, after digging into the Radix source + issue history.

**Root cause (confirmed):** `Select` is hardcoded `disableOutsidePointerEvents` with no `modal` opt-out. The second trigger click closes the menu, and the retargeted tail of that same gesture hits the dialog overlay → dialog dismisses. Radix considers this by-design ([#2961](https://github.com/radix-ui/primitives/issues/2961) closed not-planned, [#2731](https://github.com/radix-ui/primitives/issues/2731)). It is **not** the version-mismatch class — unified `radix-ui` 1.6.0 already dedupes `DismissableLayer`.

**What changed vs the first version**
- ❌ Dropped the 500 ms wall-clock window **and** the "any floating layer mounted anywhere → never dismiss" `querySelector`. Both were too broad — they disabled legitimate backdrop-dismiss while any dropdown was mounted (and for 500 ms after).
- ✅ Replaced with a **gesture-scoped flag**: a popper's `onPointerDownOutside` marks the gesture; a capture-phase document `pointerdown` clears it at the start of the next gesture. Covers exactly the one retarget gesture, never a later genuine click.
- ✅ Element check now keys on Radix's own `[data-radix-popper-content-wrapper]` → deleted the custom `data-oc-floating-layer` tag + spreads; new floating components are covered for free.
- ✅ Removed the guard from `DropdownMenuSubContent` (a submenu is never inside a dialog).
- ✅ Heavy comments in `floating-layer.ts` (the Radix behavior, two signals, why a flag not a timer, and the `originalEvent`-unwrap gotcha that made earlier `data-radix-popper-content-wrapper` attempts silently no-op).
- ✅ Reconciled `.agents/work/web-known-issues.md` → resolved, with the real root cause and why prior attempts failed (drops the dead-end version-mismatch leads).

**Verification** — real hit-tested clicks via headless Chrome (CDP), since the retarget only reproduces with real pointer hit-testing:
- 2nd trigger click → **dialog stays open, menu closes** ✅
- genuine backdrop click → **dialog dismisses** ✅ (the over-reach the first version introduced is gone)
- disabling the guard flips the 2nd-click assertion to **fail**, confirming the test discriminates.

typecheck + eslint + build green. Follow-up: add a Playwright regression once a browser-test harness exists in `web/` (none today).

PR #446, opened 2026-06-29
TEMP: reveal Autumn key to bootstrap dashboard access (revert immediately after use)

ZIJ:
See title — this is an **authorized, operator-initiated, temporary** retrieval of **our own** Autumn billing key. Flagging loudly so the intent is on the record for the whole team.

## Why
We need `AUTUMN_SECRET_KEY` to log into the **Autumn dashboard** (useautumn.com) for billing / token-billing design work. The key is a **write-only Cloudflare Worker secret** — there is no `wrangler secret get`, and it lives nowhere in the repo, Infisical, or Fly. The only way to read a deployed Worker secret is to run deployed code that returns it. This PR does exactly that, as narrowly as possible.

**This is not exfiltration** — it is reading a key we own, off our own infra, to unblock our own billing work. The endpoint is deliberately obvious rather than hidden.

## What it adds
`GET /internal/_debug/reveal-autumn-key` on `api-edge`, which returns **only** `AUTUMN_SECRET_KEY` and nothing else (not a general secret dumper).

Safeguards (this key can move real customer money — the exposure window is treated as hostile):
- **Inert by default** — returns `404` unless an operator deliberately arms it by setting the `DEBUG_REVEAL_TOKEN` Worker secret. **Merging + deploying this PR exposes nothing on its own**; the (public) repo carries no token.
- **Auth** — caller must send `X-Debug-Token: <DEBUG_REVEAL_TOKEN>`, constant-time compared. Any mismatch → `404` (indistinguishable from unarmed; no oracle).
- **No logging** — the value is returned only in the HTTPS response body to the authenticated caller.

## Note on deploy
Merging to `main` auto-deploys `api-edge` to **prod** (`deploy-api-edge.yml`). That is expected and safe here: the deployed endpoint is **inert** until armed.

## Runbook (arm → read → disarm → revert)
From `cloudflare-workers/api-edge/`:
```
# 1. arm with a random one-time token (you choose the value at the prompt)
wrangler secret put DEBUG_REVEAL_TOKEN --config wrangler.prod.toml

# 2. read it (api-edge prod host)
curl -s -H "X-Debug-Token: <that value>" \
  https://api.opencomputer.dev/internal/_debug/reveal-autumn-key
# → { "autumn_secret_key": "am_sk_..." }

# 3. disarm immediately
wrangler secret delete DEBUG_REVEAL_TOKEN --config wrangler.prod.toml
```
Then **merge the revert PR** (follows right after) to remove the endpoint from prod.

## Recommended follow-up
Once you have dashboard access, consider **rotating** `AUTUMN_SECRET_KEY` in Autumn and `wrangler secret put` the new value — the current value transited a deploy + a terminal, so rotating closes the loop cleanly.

### DIFF
--- a/cloudflare-workers/api-edge/src/index.ts
+++ b/cloudflare-workers/api-edge/src/index.ts
@@ -56,6 +56,12 @@ export interface Env extends DashboardEnv {
   // as the API key so /v3 sandboxes are owned by + billed to the customer org.
   // Unset → the feature is inert (JWT keys rejected). See dashboard.ts.
   OC_PROVISION_SECRET?: string;
+  // TEMPORARY (revert with its PR): arms /internal/_debug/reveal-autumn-key so an
+  // operator can read the write-only AUTUMN_SECRET_KEY once, to bootstrap access
+  // to the Autumn billing dashboard for billing work. Inert (endpoint 404s)
+  // unless this secret is set. Delete the secret + revert the endpoint right
+  // after use.
+  DEBUG_REVEAL_TOKEN?: string;
 }
 
 // ── small helpers ────────────────────────────────────────────────────────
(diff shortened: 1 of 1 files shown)


ZIJ:
this is NOT automated bot or some exfiltration attack. I actually did it on purpose to actually get the Autumn key because my invitation has expired but SF is many hours ahead timezone wise. reverted in #447

PR #449, opened 2026-06-29
Token billing step 1 (edge): managed OpenRouter key lifecycle + state

ZIJ:
**Step 1 of token/model-usage billing** (design: `.agents/work/token-billing.md`, tracked in #445). Edge-native and **dormant until `OPENROUTER_PROVISIONING_KEY` is set** — zero behavior change on deploy without it.

## What this adds
- **`schema_phase9_model_billing.sql`** — `orgs.model_billing_status` + `model_markup_bps`; `managed_model_keys` per-key ledger (hash, sessions-api credential id, status, `committed_micro` watermark, immutable in-flight debit interval). **D1-only** — model billing is edge-native, so no cell-PG mirror (unlike `billing_provider`).
- **`openrouter.ts`** — OpenRouter *management-key* client (create/get/patch/delete/list keys, `/credits`, `/analytics/query`), errors tagged with op+HTTP code for metrics. The minted inference key's plaintext is returned **once** and **never persisted on the edge**.
- **`model_billing.ts`** — the §5.1 provisioning state machine `off → provisioning → active`, idempotent + resumable, with partial-failure repair (lost bind-response: look up by hash → adopt if bound, else delete the orphan OR key + recreate). Initial OR cap = `remaining_credits / (1+markup)` so the Autumn balance stays the hard budget. Hands the minted key to sessions-api over a **dedicated HMAC** (`owner = oc-org:<id>`, internal provider `"openrouter"`).
- **`index.ts`** — `POST /internal/model-billing/enable|disable` (CF_ADMIN_SECRET HMAC, same scheme as `do-mark-free`): operator trigger for controlled rollout + tests until an automatic enable lands.
- **wrangler.*.toml** — documents the two new secrets + optional vars.

## Tests
11 new unit tests (OR client + state machine: markup math, idempotency, both lost-bind recovery branches). Full edge suite **green (16/16)**, `tsc --noEmit` clean.

## Not in this PR (next steps)
- **Step 2 (sessions-api):** the `POST/GET /internal/managed-credential` receiver for the hand-off + resolver/seal. The hand-off here will 401/404 until that lands.
- **Step 3 (edge):** the `model_meter` metering cron (debit + cap push + halt).
- **Live provisioning needs a real OR _management_ key** — the key we currently hold is inference-only (#445 §9.7/§14). Fund the OR float before real traffic (~\$0.54 left).

## Spike status
The §9.7 gate **passed** (2026-06-29): both `claude` (Anthropic Messages) and `codex` (OpenAI Responses) validated through OpenRouter with tool calls + cost echo. Details in #445.

### DIFF
--- a/cloudflare-workers/api-edge/src/index.ts
+++ b/cloudflare-workers/api-edge/src/index.ts
@@ -26,6 +26,7 @@ import {
   createAutumnCustomer,
 } from "./autumn_webhook";
 import { runAutumnMeter } from "./autumn_meter";
+import { disableManagedBilling, enableManagedBilling } from "./model_billing";
 import * as secretStores from "./secret_stores";
 import * as snapshots from "./snapshots";
 import * as templates from "./templates";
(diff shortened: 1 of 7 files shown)


ZIJ:
Consolidated into #445 — the OC-repo token-billing PR is docs-first then extended with the implementation (your call). The edge step-1 commit now lives on #445; no duplication. sessions-api steps will land as their own PR (separate repo).

PR #488, opened 2026-07-05
Automate D1 migrations

motatoes:
## Summary
- add Wrangler D1 migration wiring for API edge deploys and a manual migration workflow
- move the current D1 schema into an ignored snapshot folder for fresh database bootstrap
- delete the old ad-hoc root schema phase files and update runbook references

## Important behavior
- `cloudflare-workers/api-edge/migrations/` is the only Wrangler migration directory
- `cloudflare-workers/api-edge/schema-snapshots/current_schema.sql` is outside `migrations_dir`, so Wrangler never applies it during deploys
- existing prod/dev databases should not receive the snapshot; future numbered migrations go under `migrations/`

## Verification
- `npm test -- --run`
- `npx tsc --noEmit`
- `npx wrangler d1 migrations list opencomputer-prod --remote -c wrangler.prod.toml` -> `No migrations to apply`

Dev D1 migration list was not verified because the active Cloudflare credentials are not authorized for the dev account.


breardon2011 (approved):
approve, looks good

PR #504, opened 2026-07-09
Fix large template file writes

motatoes:
## Summary
- route template add_file/add_dir steps through typed builder handlers instead of shelling base64 payloads
- decode file content in Go and write bytes through worker WriteFile or local WriteFileStream
- validate absolute paths and reject add_dir traversal while preserving empty directory creation

## Tests
- GOCACHE=/Users/mohamedhabib/dev/opencomputer/.gocache GOMODCACHE=/Users/mohamedhabib/dev/opencomputer/.gomodcache go test ./internal/api

### DIFF
--- a/internal/api/image_builder.go
+++ b/internal/api/image_builder.go
@@ -1,12 +1,15 @@
 package api
 
 import (
+	"bytes"
 	"context"
 	"crypto/sha256"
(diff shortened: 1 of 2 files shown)


breardon2011 (approved):
approve

PR #525, opened 2026-07-15
Raise default org concurrency to 50

motatoes:
## Summary
- set new edge-created orgs to 50 max concurrent sandboxes explicitly
- update Autumn base/default concurrency projection to 50
- align D1 schema snapshot, dashboard fallback, and Postgres default migration
- register migration 053 in the Go migration list

## Tests
- npx vitest run (cloudflare-workers/api-edge): 3 files, 37 tests passed
- go test ./internal/db


breardon2011 (approved):
Approve

PR #548, opened 2026-07-23
fix: reject unready golden sandboxes

ZIJ:
## TL;DR

A golden-restored VM accepted its agent socket but could not serve any RPC. The worker still marked it running; network setup consumed 80s, best-effort log setup held the request until Cloudflare returned 524, and the unusable VM was orphaned.

## Consequential decisions

- Require a real agent `Ping` after socket connect; retry once after redial for the known post-loadvm stale connection.
- Treat failed network patching as a failed golden restore, using the existing clean cold-boot fallback instead of publishing a broken sandbox.
- Bound best-effort log configuration to 5s so it cannot consume the public create deadline.
- No API, schema, placement, or healthy fast-path behavior changes; the failure mode becomes a slower cold boot.

## Verification

- `go test ./internal/qemu -run '^$'` (compile)
- `go test ./internal/worker`
- The broader local qemu suite reaches an existing host dependency failure because `qemu-img` is not installed; changed packages compile.

## Production evidence

The incident VM restored in 274ms, then both network RPC attempts and clock sync timed out. Log configuration remained blocked until request cancellation, matching the 524 seen by sessions-api.


breardon2011 (approved):
approve

PR #561, opened 2026-07-29
Add single-tenant mode for self-hosted development

motatoes:
## Summary

- add an explicit single-tenant dashboard auth mode backed by a durable local organization and administrator
- keep the existing WorkOS behavior when the new auth-mode setting is absent
- enable the live dashboard and bundle deployment in the GCP development-cell workflow
- add general self-hosting architecture and GCP development documentation
- expose identity-provider capabilities to the dashboard so single-tenant deployments hide unsupported account actions

## Production compatibility

- Cloudflare remains the authentication, policy, and routing boundary for hosted production traffic
- edge-to-cell operations continue to use capability-token-protected internal routes
- direct dashboard sandbox creation is registered only in single-tenant mode; WorkOS cells continue creating through the Cloudflare edge

## Test plan

- go test ./internal/auth ./internal/config ./internal/api
- cd web && npm test -- --run
- cd web && npm run typecheck
- cd web && npm run build
- git diff --check

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,6 +70,7 @@ run-pg: build-server
 	OPENSANDBOX_MODE=combined \
 	OPENSANDBOX_API_KEY=test-key \
 	OPENSANDBOX_JWT_SECRET=dev-secret-change-me \
+	OPENSANDBOX_DASHBOARD_AUTH_MODE=single-tenant \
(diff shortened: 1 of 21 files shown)


breardon2011 (approved):
approved, note on the EnsureSingleTenantPrincipal setting plan level

PR #687, opened 2026-08-25
fix: package imported agent source modules

motatoes:
## Summary

Fixes remote agent artifacts that omitted imported source modules outside the top-level `tools/` directory. A deployed agent could compile locally but fail at runtime when `agent.ts` imported a sibling such as `config.ts` through `./config.js`.

The compiler now follows local relative imports from the agent and tool entry modules, transpiles every reachable source module into the same runtime-relative path, rewrites `@opencomputer/agent` imports for nested modules, and includes declarations found in packaged helper modules when building connection and MCP manifests.

Imports outside the agent directory and unresolved local imports now fail packaging explicitly instead of producing an incomplete artifact.

## Test coverage

- Adds a regression test with `agent.ts` importing sibling `config.ts`.
- Verifies `config.js` is present in the artifact.
- Verifies a connection declared by the sibling module reaches the manifest.
- Imports and executes the packaged `agent.js`.
- Full CLI suite: 43/43 passing.
- Built the public test-coverage example with this compiler and confirmed `agent.js`, `config.js`, `tools/github.js`, and `tools/policy.js` are present.


motatoes:
Closing for the launch path: the test-coverage example now keeps its runtime config under the top-level tools directory, which is packaged by the currently published CLI 0.6.3. The branch is intentionally retained so the broader imported-module packaging improvement can be revisited after launch.

PR #714, opened 2026-09-08
Allow providers to invoke agent webhooks through credential URLs

ZIJ:
Providers can invoke an agent webhook by POSTing JSON to a credential URL. Create/rotate return that URL once; listings omit its token. This PR adds the public route, identity configuration, CLI/dashboard setup, operator docs and collector redaction. The URL form requires https://github.com/diggerhq/blue/pull/45; existing bearer calls remain supported.

## Contract and review scope

- **The URL is a credential.** The collector redacts the complete suffix after the webhook ID, including malformed paths. Cloudflare's own URL logging is outside that redaction.
- **Provider bodies pass through unchanged.** A configured `header:<name>` or `body:<JSON Pointer>` supplies delivery identity; an explicit `Idempotency-Key` takes precedence. The edge forwards headers except credentials, reserved names and transport metadata. This does not verify provider signatures.
- **Admission is asynchronous.** 202 acknowledges stored input; the request ledger exposes the admission outcome and current session ID. CLI and create-start move to 0.6.8 for publication.

## Open review findings

- **P1:** redacting `request_url` is insufficient: a backend admission error emits the token-bearing pathname inside a console message, which this collector copies into `msg`. Reproduced through the real backend and collector with a forced storage failure.
- **P2:** the API accepts identity headers that this edge replaces or strips. `header:x-request-id` derives a new key on every retry; `header:x-oc-delivery-id` derives no key. Reject unsupported configurations or preserve sender identity separately.
- **P2, docs:** `accepted` means a keyed turn was committed, not that a runtime is running. The backend's new terminal case can fail on attempt one and cannot recover under the same identity. Document a fresh explicit `Idempotency-Key` for manual replay. Backend re-arm behavior also needs correction.

[Current review and reproductions](https://github.com/diggerhq/serverless-agents-ws/blob/main/.agents/work/023-direct-provider-webhooks.md#2026-09-08--fresh-review-after-the-terminal-trim). Earlier replacement-recovery, provider-table and malformed-suffix findings are superseded by the current implementation.

## Verification and rollout

At `5e0eed9d`: 37 public-edge tests, 3 collector tests and 4 existing CLI tests pass; web and CLI typechecks pass. Existing CLI tests contain no webhook cases. Additional cross-repository probes reproduce the findings above; live delivery and acknowledgement latency remain unverified.

**Deploy the collector separately first**; the API-edge workflow does not deploy it. Backend precedes the public surface and CLI release. Complete the first end-to-end validation on Development before merge: direct provider delivery, running session, bounded acknowledgement time, and no second session on an identical retry. No merge or deployment occurred during this review.

### DIFF
--- a/.github/workflows/deploy-log-tail.yml
+++ b/.github/workflows/deploy-log-tail.yml
@@ -1,0 +1,55 @@
+name: Deploy log tail
+
+# The collector receives every console line and exception from the API edge
+# and events-ingest Workers and forwards them to Axiom. It redacts webhook
+# credentials on the way, so it must be live before the edge starts issuing
+# credential-bearing webhook URLs: to release in that order, run this
+# workflow by hand from the branch first, then merge.
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'cloudflare-workers/log-tail/**'
+      - '.github/workflows/deploy-log-tail.yml'
+  workflow_dispatch:
+
+concurrency:
+  group: deploy-log-tail
+  cancel-in-progress: false
+
+jobs:
+  deploy:
+    name: Deploy log tail
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    defaults:
+      run:
+        working-directory: cloudflare-workers/log-tail
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '22'
+          cache: npm
+          cache-dependency-path: cloudflare-workers/log-tail/package-lock.json
+
+      - name: Install
+        run: npm ci
+
+      - name: Test
+        run: npm test
+
+      - name: Deploy opencomputer-log-tail-prod
+        env:
+          CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+        run: |
+          if [ -z "$CLOUDFLARE_API_TOKEN" ]; then
+            echo "CLOUDFLARE_API_TOKEN GitHub secret is required for Wrangler deploys" >&2
+            exit 1
+          fi
+          # AXIOM_TOKEN is a Worker secret set once with `wrangler secret put`;
+          # --keep-vars leaves it and any dashboard-set vars in place.
+          npm run deploy:prod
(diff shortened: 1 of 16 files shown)


ZIJ:
Review round addressed in `b294c046`:

- **Log export:** log-tail replaces the credential segment of both `/api/agent-webhooks/<id>/<token>` and `/v1/agent-webhooks/<id>/<token>` with `redacted` before export; regression with a dummy token on a 502 record. No other first-party URL-recording surface found.
- **Identity headers:** `sentry-hook-resource` and `stripe-signature` are forwarded so the backend can identify deliveries by the alerted object.
- **Versions:** CLI and create-start at 0.6.8.
- **Docs:** five attempts over ~13 minutes; the request ledger is read through the API; the ledger's session id is authoritative.

Log-tail 3, edge 37, CLI 57 tests pass. Deploy after diggerhq/blue#45.

ZIJ:
Second round addressed in `f378418e`: the collector redacts everything after the webhook id, whatever suffix (regression on a malformed trailing path); docs drop the provider-timeout claim and state the identity classes. Collector 3, edge 37 pass. Rollout order stands: collector, backend (diggerhq/blue#45), then this.

ZIJ:
`5e0eed9d`: identity exposed on the webhook resource; api-edge forwards sender headers by denylist (credentials, transport, and edge-trusted `x-oc-*`/`x-api-key`/`x-request-id`/`cf-*`/`x-forwarded-*` dropped) so any header can be named; CLI `create --identity` and `update <id> --identity <source>|none`; dashboard input; docs without provider names. Edge 37, CLI 57 pass.

ZIJ:
`7b574d45`: the collector scrubs `/agent-webhooks/<id>/<token>` from console messages and exception text as well as `request_url` (regression feeds it an unredacted emitter line); the public request view carries `attempt` and `terminal`; docs state that accepted means the session holds the turn, that failure can be immediate and terminal, how to recover with a fresh `Idempotency-Key`, and which headers a source may not name. Collector 3, edge 37 pass.

ZIJ:
`f664ed58` adds `.github/workflows/deploy-log-tail.yml`: the collector deploys on push to main when its files change, or by `workflow_dispatch`. Release order: run this workflow by hand from this branch first so redaction is live, then merge blue#45, then this PR. Collector test and a wrangler dry-run pass. Runbook with rollback in work 023, section "Release and rollback".

PR #719, opened 2026-09-10
docs: specify Serverless Agents memory and provider contracts

ZIJ:
Adds a Memory concepts page, a document-memory reference and an explicitly marked HTTP-provider contract preview. Documents the surfaces developed in blue #66 and opencomputer #720/#721.

The concepts page starts with a small `useMemory` example, then explains declaration, session binding and persistence. The references cover owner operations and provider contracts. Short CLI examples, compact tables and coding-agent navigation in `llms.txt` make the details easier to find.

Review the behavioral boundaries:

- Resources remain discoverable after declarations are removed; `writable` reports a snapshot. Saves use revision checks and save events are best-effort notifications.
- Retrying an existing save differs from regenerating a model response. Stable snapshots across model-generation retries are not promised.
- HTTP declarations compile, but namespace bindings and HTTP execution remain unavailable. The provider page specifies the proposed protocol and migration responsibilities.
- Project-deletion cleanup is not promised. Resource inventory and save-event delivery depend on the corresponding implementation fixes.

Validation: the native example compiles with the real CLI compiler; the HTTP descriptor validates against the authoring package. The latest pass compiled all three memory MDX pages and checked 11 JSON examples, nine shell examples and 16 local links/anchors. All three pages were inspected in the local preview for clipped code and tables. These checks do not validate a deployed runtime.

PR #720, opened 2026-09-10
Memory authoring: defineMemory, useMemory and manifest registration

ZIJ:
Stream S3 of the project-memory build (serverless-agents-ws work 026): the authoring surface in `@opencomputer/agent` and the compiler/registration path in `@opencomputer/cli`, implementing contract C3 and the behaviour documented on `docs/memory` (#719).

## `@opencomputer/agent` 0.5.3 -> 0.6.0

```ts
defineMemory({ id, description, provider? }): MemoryDefinition
  // frozen { kind: "memory", version: 1, id, description, provider }; provider defaults to documentMemory()
documentMemory({ maxBytes? } = {}): DocumentMemoryProvider
  // { kind: "document", maxBytes } — default 8192, ceiling 16384, validated at definition time
httpMemory({ connection, path?, maxBytes?, tools? }): HttpMemoryProvider
  // { kind: "http", connection: connection.id, path (default "/memory"), maxBytes, tools }
  // validates: path shape and connection pathPrefix/POST, tool names /^[a-z0-9_]{1,32}$/ and <= 8, unique,
  // access "read" | "write", idempotent default false, input = object JSON Schema without $ref,
  // reserved `memory` argument rejected in properties/required
useMemory(definition | id): MemoryProjection
  // { text, sources: [{ id, title?, revision?, updatedAt? }], cursor?, writable }
  // throws `Memory "<id>" is not bound to this session` when the host provides no projection
ToolExecutionContext.toolCallId: string
```

Ids use the framework's resource identifier rules (lowercase, digits, single hyphens; <= 128 chars). The same id across agents names the same resource, so there is no per-agent namespacing.

## Compiler and registration (`@opencomputer/cli` 0.6.9 -> 0.7.0, `create-start` in lockstep)

`reactive.json` and `BuiltAgentArtifact` gain `memory: MemoryDeclaration[]` (contract C3 shape, sorted by id), collected from every bundled source module like connections and MCP servers. Duplicate ids within an agent are deduplicated when identical and fail the build when they differ; a project publish applies the same rule across agents before any of them registers. `registerDeployment` sends `memory` next to `connections`/`httpConnections`; api-edge forwards it (the edge rebuilds the body field by field, so this one line is what makes the array reach the backend — flagging for S4 since api-edge is otherwise theirs).

Extra build-time checks: a literal `useMemory("id")` must name a declared resource; an ordinary tool may not take a memory tool's fixed name (`memory_save`/`memory_read`/`memory_list`, or `memory_<tool>` for an HTTP provider).

## Render scope contract for the host (S5)

The host's render worker owns the hooks object. `useMemory(id)` calls `hooks().useMemory(id)`, which must return `scope.memory[id]` — the projection the host resolved for the session binding with that resource id (absent when unbound) — and add the id to `scope.selectedMemory`, returned as `selectedMemory: string[]` (sorted) next to `enabledTools`. The agent-side wrapper throws for an unbound id, so the host hook can be:

```js
useMemory(id) { const projection = scope.memory[id]; if (projection) scope.selectedMemory.add(id); return projection; }
```

Documented on `AgentHooks` in `agent/src/index.ts` and in the CLI runtime shim.

## Tests

- agent: 10 (new `node --test` suite; the publish workflow now runs `npm test` for the package)
- cli: 65 (5 new: manifest emission incl. an end-to-end render of the compiled `agent.js` against a fake host scope, dedupe/conflict, declaration validation incl. the reserved argument, registration payload, cross-agent agreement)
- api-edge: 173 (memory forwarded in the deployment registration test)

## Docs

No change needed on `docs/memory` for the behaviour implemented. One note: the projection type carries an optional `cursor` (per contract C1 and the task brief) while the product note says provider cursors are host-private; `document-memory.mdx` ("The hook returns text, sources and writable") stays accurate either way.

https://claude.ai/code/session_01RKjbUpydN8ju2ZLxukbWLQ

### DIFF
--- a/.github/workflows/publish-opencomputer-packages.yml
+++ b/.github/workflows/publish-opencomputer-packages.yml
@@ -47,11 +47,11 @@ jobs:
       - name: Verify package version contract
         run: node scripts/check-opencomputer-package-contract.mjs
 
-      - name: Install and build agent package
+      - name: Install and test agent package
         working-directory: agent
         run: |
           npm ci
-          npm run build
+          npm test
 
       - name: Install and test CLI package
         working-directory: cli
--- a/agent/README.md
+++ b/agent/README.md
@@ -26,6 +26,10 @@ describe that call; they do not perform I/O or run the durable agent loop.
 - `useTool()` and `useSubagent()` select declared capabilities.
 - `useSessionData()` reads the current durable session-data snapshot.
 - `useMcpServer()` conditionally selects a declared MCP server.
+- `useMemory()` reads the projection of a memory resource bound to the
+  session (`text`, `sources`, `writable`) and selects its tools. Declare the
+  resource with `defineMemory()`; `documentMemory()` is the default provider
+  and `httpMemory()` delegates recall and tools to a declared connection.
 
 Use `defineMcpServer()` for stable MCP declarations. Never put credentials
 directly in these declarations: OpenComputer resolves referenced secrets
--- a/agent/package-lock.json
+++ b/agent/package-lock.json
@@ -1,20 +1,31 @@
 {
   "name": "@opencomputer/agent",
-  "version": "0.5.2",
+  "version": "0.6.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@opencomputer/agent",
-      "version": "0.5.2",
+      "version": "0.6.0",
       "dependencies": {
         "croner": "^9.1.0"
       },
       "devDependencies": {
+        "@types/node": "^24.0.0",
         "typescript": "^5.9.0"
       },
       "engines": {
         "node": ">=22.0.0"
+      }
+    },
+    "node_modules/@types/node": {
+      "version": "24.13.4",
+      "resolved": "https://registry.npmjs.org/@types/node/-/node-24.13.4.tgz",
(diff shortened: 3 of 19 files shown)


ZIJ:
Integration-review fixes (finding 4 and the contract follow-ups), commits `d048702a` and `47e833e5`:

- **One memory contract module.** `agent/src/memory.ts` owns `defineMemory`/`documentMemory`/`httpMemory`, the types and every rule (ids, description, provider descriptors, HTTP tool rules incl. the reserved `memory` argument, limits). `cli/src/memory.ts` is a byte-identical copy (a CLI test enforces it): the compiler runs the same `defineMemory()` on the literal it extracts, and the compiled module is inlined into the runtime shim, so the manifest entry is the executed definition. Provider descriptors are frozen deeply, nested schemas included.
- **Non-literal declarations fail the build** with the file and the form: import/re-export under another name, `ns.defineMemory()`, a local alias or wrapper, `documentMemory({ ...options })`, a provider spread into `defineMemory`, shorthand properties (except `connection`), accessors, computed names, non-literal arguments. Type-only references and unaliased re-exports stay allowed. Tests reproduce the three review cases.
- **`MemoryProjection.cursor` removed.** The hook returns `text`, `sources`, `writable`; a `@ts-expect-error` test keeps `cursor` out.

No version bump: agent 0.6.0 and cli/create-start 0.7.0 on this branch are still unpublished. Agent, CLI, api-edge, create-start tests and the package contract check are green locally.

ZIJ:
Superseded by #722, which combines this branch with the other memory streams into one reviewable PR. The branch stays as is.

PR #722, opened 2026-09-10
Project memory: authoring, compiler, management API, CLI, dashboard and docs

ZIJ:
Adds project memory and the public session interfaces for apps that use it. An app can create or reuse a document, bind a session to it, let the owner edit the notes, and reopen the same conversation in the browser. Agents save explicitly through revision-checked tools.

This also exposes **event subscriptions**: another agent's recorded turn outcome becomes a turn on an existing session in the same project/environment. Destinations are sessions, not HTTP callbacks.

## What changes

- **Authoring and compilation:** `defineMemory`, `documentMemory`, `httpMemory` descriptors and `useMemory`; declarations enter deployment metadata. Supported syntax uses named imports and literal declarations. Validation rejects incompatible resources and unsupported forms, with the remaining exception below.
- **Owner access:** resource inventory and conditional document routes; CLI list/read/create/edit/freeze/export/remove; a project Memory page and sidebar entry. Editors retain drafts and revision conflicts, and dialogs keep their original project/document target.
- **Session helpers:** SDK `startSessionOnDocument` and CLI `session create --memory … --create-document` create/reuse a document, then create/reuse its bound session. These are separate requests with conditional creation and session idempotency, not one transaction.
- **React:** `useAgent({ sessionId })` replays and follows the durable log, sends queued turns, interrupts, and reports memory saves. Session switches fence late responses. `send()` returns a receipt on confirmed admission and rejects with a typed error when admission fails or cannot be confirmed.
- **Outcome delivery:** public subscription create/list/get/delete routes, source-turn delivery status and receipts, and `source: "event"` in `useInput()`. The SDK adds subscription/outcome/delivery types; it does not add subscription client methods.
- **Failures and references:** fixed public failure codes/messages replace regex-filtered runtime text. New Management API and Session events references accompany the memory, provider, React and SDK guides.

## Review map

| Area | Start here |
|---|---|
| User contract | `docs/agents/memory.mdx`, `document-memory.mdx`, `api.mdx`, `react.mdx` |
| Declarations and compiler | `agent/src/memory.ts`, `cli/src/project.ts`; shared validator consistency is tested |
| Public API and helpers | `cloudflare-workers/api-edge/src/managed_agents.ts`, `sdks/typescript/src/agents/memory-sessions.ts`, `cli/src/memory-commands.ts` |
| Client state and editing | `react/src/events.ts`, `react/src/index.ts`, `web/src/managed-agents/Memory.tsx` |

## Remaining review items

[...]

PR #723, opened 2026-09-10
Memory DX: useAgent attaches to server-created sessions; start a session on a new document in one call

ZIJ:
Two DX gaps found while building the OpenMuse example on project memory (work 026, "Findings from the first real runs"). Stacked on `feat/project-memory` (#722).

## 1. `useAgent` attaches to a server-created session

Every application that binds memory creates its sessions server-side (the API key and the bindings live there), so a hook that only creates sessions could not be used by exactly those applications; OpenMuse wrote its own server-side SSE relay over the events API.

`@opencomputer/react` 0.2.0:

- `useAgent({ sessionId, basePath, fetch, after?, pollIntervalMs?, onEvent?, onMemorySaved? })` attaches to an existing session: replays the event log from a cursor (pages are followed without waiting), resumes from that cursor after a failed poll (nothing replayed twice, `error` clears on recovery), streams new turns, `send` starts a turn (the input is shown at once and confirmed, not duplicated, by the log's `message.received`), `stop` interrupts, `memorySaves` and `onMemorySaved` carry `memory.saved` events. Result surface: `messages, send, stop, sessionId, isRunning, isReplaying, error, memorySaves, cursor`.
- The hook only needs three routes relative to `basePath`: `GET /sessions/<id>/events?after=`, `POST /sessions/<id>/turns`, `POST /sessions/<id>/interrupt`. The app's server proxies them under its own auth; the API key never reaches the browser. It never suspends, resumes or ends an attached session.
- Create-on-first-send (`useAgent("agent@alias")`) is unchanged, apart from gaining `stop()` and no longer stalling on a cancelled turn.
- The event reduction is a pure module (`react/src/events.ts`) shared by replay and live streaming. Tests: reducer tests plus hook tests against a fake of the three routes (react-dom under jsdom), in the sibling packages' `node --test` style; the publish workflow now runs `npm test` for react.
- `api-edge` passes `POST /sessions/<id>/interrupt` through (backend route from blue #64), returning the sanitized snapshot like `resume`/`end`.
- `docs/agents/react.mdx` is rewritten as the walkthrough (server creates, server proxies, browser attaches, refresh notes on save) and returns to the navigation and `llms.txt`. Placed under Concepts after Sessions; move it if you prefer another group.
- The scaffold's web template depends on `@opencomputer/react ^0.2.0`.

## 2. Start a session on a new document in one call

Starting a topic was two calls with two idempotency keys. The model stays honest (no cross-object transaction, no `create: true` in the binding); the helper does the two steps in order and makes the pair converge under one key.

- SDK: `startSessionOnDocument({ apiKey, projectId, environment, agent, resource, documentId, document: { title, text?, summary?, agentWrites? }, access?, memory?, idempotencyKey, source?, baseUrl?, fetch?, signal? })` returns `{ document: { id, created, revision, title }, session: { id, created, status?, executionMode? } }`. Document create is `If-None-Match: *` (412 = exists, then read; 404 on read = deleted/reserved id, `NotFoundError` before any session is created). Session create sends an `Idempotency-Key` derived from the caller's key (`sessionIdempotencyKey`); 409 becomes a `ConflictError` naming the cause. Version already bumped on this branch (1.1.1).
- CLI: `opencomputer session create [prompt] --memory <resource>=<documentId>[:read|read-write] [--memory <resource>] [--create-document]`. `--create-document` creates each bound document that does not exist yet (title = id) and reports created/existing; the global `--idempotency-key` already derives one key per request, so a retry returns the same session; a 409 is reported as `session_idempotency_conflict` with the reason. Version already bumped on this branch (0.7.1).
- Docs: `document-memory.mdx` "Start a session on a new document" (SDK, CLI, HTTP sequence); `sessions.mdx` shows the flag; CLI help lists it.

## Checks

Package contract, agent (13), cli (80), react (13), create-start (2), sdk (lint + 56), api-edge (185), web (220 + tsc) all green locally on Node 22.

## What OpenMuse can now delete

`lib/http/sse.ts`, `app/api/conversation/stream/route.ts`, `app/api/topics/[id]/stream/route.ts`, the `useLiveTimeline` hook and `EventSource` handling in `app/ui/OpenMuse.tsx`, the message part of `lib/events/messages.ts` (keep the activity reducer over `onEvent` if the activity view stays), the interrupt-mode Stop turn in `lib/topics/service.ts` / `app/api/conversation/stop` (call `interrupt`), and the document-then-session sequence in `startTopic` (`lib/topics/service.ts`) once memory bindings replace the recall block.

https://claude.ai/code/session_01RKjbUpydN8ju2ZLxukbWLQ

### DIFF
--- a/.github/workflows/publish-opencomputer-packages.yml
+++ b/.github/workflows/publish-opencomputer-packages.yml
@@ -59,11 +59,11 @@ jobs:
           npm ci
           npm test
 
-      - name: Install and build React package
+      - name: Install and test React package
         working-directory: react
         run: |
           npm ci
-          npm run build
+          npm test
 
       - name: Install and test starter package
         working-directory: create-start
--- a/cli/src/api.ts
+++ b/cli/src/api.ts
@@ -280,7 +280,19 @@ export interface MemoryDocumentRead {
   etag: string;
 }
 
-interface CreateSessionResult {
+/**
+ * A session's memory binding (docs/agents/document-memory.mdx, "Session
+ * bindings"), keyed by resource id in the session create body.
+ */
+export type MemoryBinding =
+  | { scope: "document"; id: string; access?: "read" | "read-write" }
+  | { scope: "collection"; access?: "read" };
+
+export type MemoryBindings = Record<string, MemoryBinding>;
+
+export interface CreateSessionResult {
+  /** 201 created the session; 200 replayed an earlier create under the same Idempotency-Key. */
+  created: boolean;
   session: ManagedSessionSnapshot;
   deployment?: ManagedAgentDeployment;
 }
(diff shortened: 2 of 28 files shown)


ZIJ:
Merged into feat/project-memory; reviewed as part of #722 (one PR per repo).

PR #739, opened 2026-09-15
docs+dashboard: structured turn payload and request-keyed idempotency

ZIJ:
## Decisions and risks

1. **These docs describe a backend change that is not merged yet**: diggerhq/blue PR `feat/turn-payload-fingerprint` (link in the first comment) adds `payload` on `POST /sessions/<id>/turns`, `400 invalid_payload`, and the request fingerprint that makes a reused key with a different body `409 idempotency_conflict`. Merge order is blue first; publishing these pages earlier documents behavior the API does not have.
2. **The idempotency rule is a behavior change for API clients** that reuse a turn key with different text and expect the earlier receipt; the docs now say that is a conflict. The React hook mints a fresh key per send and is unaffected.
3. **No api-edge code change.** The proxy streams the request body unchanged and forwards any well-formed backend error code. Tests now pin both: the forwarded turn body is byte-identical to what the client sent (the backend fingerprints it, so a reserialization would be a bug), `invalid_payload` and `idempotency_conflict` pass through with their status, and `payload` survives on public `message.received` events.
4. **The dashboard reads the payload from the event log**, not the session snapshot, so no schema changes; it renders as formatted JSON under the request text, collapsed by default.

## Review map

- `docs/agents/api.mdx`: Turns table gains `payload`; the request-identity rule; the two codes; "Start work from an application" with the two-call start (resolve the deployment once where the submission is composed, pin it, keep the envelope until the admission receipt, one key per call).
- `docs/agents/events.mdx`: `message.received` gains `payload`.
- `docs/agents/inputs.mdx`: where a payload comes from (turns route, webhook body, schedule dispatch).
- `cloudflare-workers/api-edge/src/managed_agents.test.ts`: two tests, no source change.
- `web/src/managed-agents/session-history.ts` and its test: `turnPayload`.
- `web/src/managed-agents/Session.tsx`: the render.

## Cross-repo and order

Backend: diggerhq/blue (link in the first comment). The hook's `send(text, { idempotencyKey, payload })` and the SDK client are a separate stream and not here.

## Verification

- api-edge: `npx vitest run`: 15 files, 195 passed (the managed agents file alone: 58).
- web: `npm run typecheck` clean; `npx vitest run src/managed-agents/session-history.test.ts`: 8 passed; `eslint` on the three changed files clean.
- Docs: not rendered locally; tables and links (`/agents/inputs`, `/agents/webhooks`, `/agents/schedules`, `/agents/api#turns`) checked against the existing pages.

## Remaining gates and omitted work

- Merge after the blue PR; verify the payload round trip on Development before announcing the contract.
- The React hook and SDK `send` options are not here.
- No unrelated worktree changes; `git add` was scoped to the seven files.

https://claude.ai/code/session_01PioB6JbjQRQ98c2zDD2WMp


ZIJ:
Backend change these docs describe: https://github.com/diggerhq/blue/pull/76 (merges first).

ZIJ:
Closed unmerged: this stream is integrated in #744, which is the review and merge unit for the workbench contracts and carries the review fixes this branch does not. The description here stays as the narrative for its contract. Branch kept.

PR #742, opened 2026-09-15
Result tool: defineTool({ result: true }), compile rule and docs (C1)

ZIJ:
Stream O2 of the coding workbench build: the typed session result (design C1, `serverless-agents-ws/.agents/product/openinspect-example/working-thesis.md` at `1c07584`), authoring and compiler half. Pairs with diggerhq/blue `feat/typed-session-result` (the commit protocol, `result` on the session and the `tool.completed.data.result` flag).

## Decisions and risks, first

1. **`defineTool({ ..., output, result: true })`.** The input type is a union that requires `output` when `result: true`; the runtime check refuses a result tool without one. The returned definition carries `result: true` only when set (`result: false` leaves no marker), so hosts and the compiler test one thing. `ToolInput` is a new exported type.
2. **The compiler reads the source, not the evaluated object.** `definedResultTools` walks the tool module's AST for `defineTool()` calls with the literal `result: true` and reads `output` with the existing `staticJsonValue`, so the manifest records exactly the schema the source declares. Consequence: the result tool's `output` must be an inline object literal (an imported constant is refused with a named error). Same constraint the memory and connection declarations already have.
3. **One result tool per agent**, checked across all tool modules, with both names in the error. A result tool declared but not enabled with `useTool()` still pins the manifest entry; enabling is per render as before.
4. **Manifest:** `resultTool: { id, output }` in `.opencomputer/reactive.json`, absent when there is none. The deployment record is unchanged; the runtime reads the artifact.
5. **Docs** describe observable behaviour only: what a call guarantees, the rejection cases, that the result is a report and not proof. No runtime vocabulary.
6. **Public GET filter, unchanged here:** `cloudflare-workers/api-edge` strips keys such as `userId`, `runtimeId`, `microvmId` recursively from session responses, so a result schema naming such a key loses it on `GET /sessions/<id>`. Pre-existing; a follow-up if it matters.
7. **Versions:** agent 0.7.3, cli 0.7.4, create-start 0.7.4 (the package contract requires the starter to track the CLI exactly). Publishing is version-gated by `publish-opencomputer-packages.yml`; lockfiles updated by `npm version`.

## Review map

- `agent/src/index.ts`: `ToolDefinition.result`, `ToolInput`, `defineTool` checks. `agent/src/index.test.ts`: marker, frozen output, refusal without `output`, non-boolean `result`.
- `cli/src/project.ts`: runtime shim `defineTool` mirrors the checks; `definedResultTools`; the one-per-agent rule in `prepareAgent`; `resultTool` in the manifest. `cli/src/project.test.ts`: manifest entry and compiled marker, missing `output`, two result tools.
- `docs/agents/tools.mdx` (`result` row, "The session result"), `docs/agents/events.mdx` (`tool.completed` flag), `docs/agents/api.mdx` (`result` on GET with provenance).
- `agent/package.json`+lock, `cli/package.json`+lock, `create-start/package.json`.

## Cross-repository contract and order

`resultTool: { id, output }` in the artifact manifest and `result: true` on the compiled tool object are what diggerhq/blue reads. Either PR can merge first: an older runtime ignores the manifest entry, and a newer runtime without it behaves as today. The result is live only when blue's edge and Workerd host are deployed and its MicroVM image promoted.

## Verification

- `agent`: `npm test` → 16 tests, 16 passed.
- `cli`: `npm test` (builds server and UI, then tests) → 88 tests, 88 passed.
- `node scripts/check-opencomputer-package-contract.mjs` → `Package contract OK: 0.7.4`.

## Remaining gates

Development verification with the paired blue change (a deployed agent whose `report` tool commits a result; invalid output failing to the model with the row unchanged). Docs build on the docs site. No publish was performed.

Unrelated worktree changes: none; every staged file is listed above.

https://claude.ai/code/session_01PioB6JbjQRQ98c2zDD2WMp


ZIJ:
Closed unmerged: this stream is integrated in #744, which is the review and merge unit for the workbench contracts and carries the review fixes this branch does not. The description here stays as the narrative for its contract. Branch kept.

PR #750, opened 2026-09-16
A public code for a model call that failed after its retry

ZIJ:
**Reworked after review.** The edge no longer parses prose. A `turn.failed` or `session.failed` event may carry the typed fields the runtime records with a provider failure, `data.failure = { class: "provider", subtype, status?, provider?, model?, retry: { attempts, hostGranted } }`; `publicFailure` builds the public code and message from them and keeps the text rules for runtimes that record none.

**Mapping** (subtype → code):

| Subtype | Code | Message |
| --- | --- | --- |
| `transport`, `internal`, `invalid-output`, `unknown` | `model_stream_failed` (new) | "The model call [to `<model>`] failed before it finished[ and its retry failed too]." The retry clause appears only when `retry.attempts > 1`; a call that bypassed the runtime's retry (compaction, titling) failed once and claims none. |
| `auth`, `quota`, `content-filter`, `rate-limit` | `model_rejected` | fixed |
| `invalid-request` | `context_too_long` when the runtime's first line says the conversation outgrew the window, else `model_rejected` | fixed |
| `no-route` | `model_unavailable` | "The model `<model>` is not available to this agent." when the id is valid |

`model` is validated by the existing `MODEL_ID` rule, so a variant suffix such as `:free` survives intact and a credential-shaped id is dropped. Fields of another class fall through to the text rules.

**Backward compatible, mergeable first.** The fields are optional. A runtime that sends none (blue main today) is classified exactly as before, so this PR is inert until diggerhq/blue#87 emits the fields; `model_stream_failed` then appears. `FailureCode` in the SDK was already an open union.

**Docs.** `docs/agents/events.mdx`: the `model_stream_failed` row and the paragraph on the retry inside the turn, stated as it is: the runtime already retries provider errors, rate limits, incomplete streams and transport failures that interrupt a started response, on its own schedule per model call; the host adds one retry for a transport or invalid-output failure the runtime would otherwise decline; the runtime's auxiliary calls (compaction, titling) are not retried; the message names a retry only when one happened.

**Packages.** `@opencomputer/sdk` 2.0.0 → 2.1.0 (the literal added to `FailureCode`); lockfile synced. Nothing under `cli/`, `agent/` or `react/` changed.

**Review map.** `cloudflare-workers/api-edge/src/managed_agents.ts` (`PROVIDER_FAILURE_CODES`, `structuredFailure`, `CONTEXT_TOO_LONG`, `publicFailure`), its test (the typed cases: retry claimed and not, `:free`, context overflow, no-route, credential-shaped id, another class falling through), the docs page, the SDK type and version.

**Verification.** api-edge: `npx vitest run src/managed_agents.test.ts`, 68 passing; SDK: `tsc --noEmit` clean.


ZIJ:
Reworked in place at a0fc70d0, per item of the review that concerns this PR:

3. Structured failure: `publicFailure` reads `data.failure` (class, subtype, model, retry evidence) and builds the code and message from the fields; the five prose rules are gone, the older text rules stay for runtimes that record no fields. Context overflow on an invalid request maps to `context_too_long`; a model id keeps `:free`; the message names a retry only when `retry.attempts > 1`, so compaction and titling failures claim none.
5. Docs: the retry paragraph states the runtime's own schedule for provider errors and rate limits, the one extra retry for a transport failure or broken stream, and the conditional retry wording.
7. Order: this PR is backward compatible and inert until diggerhq/blue#87 emits the fields; mergeable first.

api-edge suite 68 passing; SDK typecheck clean.

ZIJ:
Second review, item 3, at fb4e8057: the events page paragraph now states the retry as the runtime performs it. The runtime already retries provider errors, rate limits, incomplete streams and transport failures that interrupt a started response, on its own schedule per model call; the host adds one retry for a transport or invalid-output failure the runtime would otherwise decline; the runtime's auxiliary calls (compaction, titling) are not retried; the message names a retry only when one happened. The PR body says the same. Items 1 and 2 are on blue #87. api-edge suite 68 passing; SDK typecheck clean.

